PR #3, opened 2016-03-14
Added --defaults-file to mydumper and myloader

cezmunsta:
Added in the `--defaults-file` option to mydumper and myloader, which helps when using a remote backup server that connects to _n_ database servers that require a different configuration


toontong:
waiting for it. pls~

PR #15, opened 2016-08-31
Adding an option in order to not disclose the password on the command line

josev814:
I added a getPassword file that will prompt the user for a password.  The user can use -a or --ask-password to invoke the option.  If -p is set with a password, -p takes precedence


maxbube:
Hi, 

Thank you!! I will review and merge this as soon as I have time.

sebastienbarre:
That would be useful indeed!

maxbube:
Hey @nkadel-skyhook just an idea before testing and merging. Can we do this directly if -p is empty?

nkadel-skyhook:
"mysqldump" behavior if "-p" is empty is to prompt for a password. I'm not sure if you'd like to mydumper to to use that, or not.

Issue #32, opened 2017-01-10
gtid_current_pos empty.


ragboy:
using a galera cluster, I notice that my master metadata doesn't always contain GTID which is a problem. Doing some research I found this:

https://jira.mariadb.org/browse/MDEV-10279

looking at this issue, it would seem for galera clusters, wouldn't it be better to use gtid_binlog_pos? And would that need to only for galera, or maybe best to use that on mariadb 10+ and galera? Or does some type of galera detection need to be done and use binlog_pos?

```
mysql_query(conn,"SHOW MASTER STATUS");
	master=mysql_store_result(conn);
	if (master && (row=mysql_fetch_row(master))) {
		masterlog=row[0];
		masterpos=row[1];
		/* Oracle/Percona GTID */
		if(mysql_num_fields(master) == 5) {
			mastergtid=row[4];
		} else {
			/* Let's try with MariaDB 10.x */
			mysql_query(conn, "SELECT @@gtid_current_pos");
			mdb=mysql_store_result(conn);
			if (mdb && (row=mysql_fetch_row(mdb))) {
				mastergtid=row[0];
			}
		}
	}
```

maxbube:
Fixed on 0.9.3

PR #33, opened 2017-01-10
resolve issue #32 with empty GTID in master metadata on galera cluster.

ragboy:
See issue 32.

Thanks for the great work with mydumper.


maxbube:
Hi ragboy,

I will test and merge as soon as I can.

Thanks!!!

PR #36, opened 2017-01-25
Support SSL MySQL connections

### LINKED ISSUE #31: SSL connection Error
I am using ubuntu 16.04, clean install running apt-get 0.9.1 and then also tried compiling 0.9.2 from here. I get this error:

** (mydumper:13381): CRITICAL **: Error connecting to database: SSL connection error: unknown error number

If I use the mysql or mysqldump client, I have no issues, and can use --ssl=1 or --ssl=0 without a problem. I don't see any other information to help me track this down.

Thank you.

insom:
I believe that these commits will fix #31 by adding optional SSL support to `mydumper` and to `myloader`.

I also moved some connection establishment code which is used in more than one place into a common function in a common file.

This was tested locally on Linux with the Oracle MySQL client libraries.


maxbube:
Hi insom, 

This is great, I will perform some tests before merging this or give you feedback.

Thanks!!!!

nkadel-skyhook:
I'd much appreciate this support as well.

ChristopherSchultz:
+1 for this enhancement

jairud:
Hi. I see that is merged in master.

But how can I use?

I tried with params `--ssl 1` or   `-s 1` and didn't work
`--help` don't show any information about ssl

I'm using version bb963448d1183e933cc540b52a62a1da27a92117

insom:
Hi @jairud since this was merged there were commits which make building it conditional: https://github.com/maxbube/mydumper/commit/97438e2d0a078f3d2dd48ed992fb6a563fa5c337

I haven't built it to verify, but looking at the code you should see an entry in `--help` if your binary has SSL support built in.

PR #38, opened 2017-02-09
Add basic timing message per table restore

keelerm84:
This is a proof of concept for a modification I'd like to make. I'd like to get feedback on changes you would want to see made -- would you prefer this being a option to the loader, using a different timing method, is this a change you even want to support or is it fine as is?

I appreciate your time and the effort you put into this tool.


maxbube:
Hi, 

It would be nice to have something like this, but let me add some notes here:
- as myloader is multithreaded the table restore timing will vary depending on the threads running, so this time might not be really be useful.
- take in mind that your implementation is not taking care of chunks. 

I will review better later and give more feedback.

Thanks for helping to improve this tool!!!!

PR #65, opened 2017-10-05
Avoid MyISAM tables as temporary view tables

sbktrifork:
Creating MyISAM tables on galera isn't a good idea and can result in errors. 
Why not simply use the default engine instead?


maxbube:
Hi @sorenk dropping innodb tables can be expensive or even cause outages on some environments, I can't change this behavior by default

sbktrifork:
Sorry for not getting back - I think it should at least be possible to choose another engine than MyISAM... What about adding a command-line parameter for this?

PR #69, opened 2017-10-15
Dump procedures and functions using case sensitive database name

kvasserman:
Since database names are case sensitive, in a rare case when there multiple databases with names different only in case, mydumper backed up procedures and functions from multiple databases, instead of the one being backed up.


maxbube:
Hi @kvasserman that depends on lower_case_table_names variable, and the code need to take that in account. Would you be able to add that?

davidducos:
This needs to be reviewed and remerge.

davidducos:
According to https://dev.mysql.com/doc/refman/8.0/en/cast-functions.html#operator_binary, binary will be deprecated.

davidducos:
BTW, this is still an issue with mysqldump: 
```
root@ubuntu-focal:~/tablenames/forth/mydumper# mysqldump -R Test > /dev/null
root@ubuntu-focal:~/tablenames/forth/mydumper# mysqldump -R TesT > /dev/null
mysqldump: Couldn't execute 'SHOW CREATE FUNCTION `count_perftest_higher_than`': FUNCTION count_perftest_higher_than does not exist (1305)
root@ubuntu-focal:~/tablenames/forth/mydumper# mysqldump --version
mysqldump  Ver 10.13 Distrib 5.7.33-36, for debian-linux-gnu (x86_64)
```

davidducos (dismissed):

davidducos (approved):

PR #73, opened 2017-11-06
Changes to syntax to to make it more like Markdown

connerbw:
Following up on https://github.com/maxbube/mydumper/pull/59
In addition to the file extension, these are minor changes to the syntax so that the GitHub page looks nicer.


connerbw:
Duplicated by: https://github.com/maxbube/mydumper/pull/64

PR #75, opened 2017-11-15
fix bug in create_databse function test database schema file exist

wshn13:
fix bug in create_databse function test database schema file exist
just by filename which mean the **CURRENT WORK DIR**, but it should
be the file in **data directory** passed by -d


wshn13:
If I run myloader outside the data dir passed by -d, the g_file_test in create_database(line 328) will always **return false** even though the file in data dir dose exist.This will result myloader just execute "CREATE DATABASE " by default and the database charset info will lost.

maxbube:
Hi @wshn13 , 

I will test this code and merge for next release, thank you!!

PR #81, opened 2017-11-27
Fix compilation with older MySQL versions.

waptaff:
Commit 37aa06686dc9ca97abbb335f8bd4b354a9694d7c broke compilation under
at least Debian 9 and CentOS 6, as MYSQL_TYPE_JSON is not defined in the
MySQL versions they ship.


npinchot:
I was having issues under Ubuntu 14 as well. This patch worked. Thanks!

maxbube:
Hi, 

Is this a problem if you compile static against mysql 5.7 libs? I will test that and see what to do aboout this.

Thanks!!

PR #90, opened 2017-12-21
Do not check for long running queries when no-lock is on

tanji:
When using `--no-lock` option, SHOW PROCESSLIST was still checked for long running queries, in order to avoid blocking `FLUSH TABLES WITH READ LOCK` statement. However with `--no-lock` there is no FTWRL. That check is unnecessary and caused backups where `no-lock` operation is desirable to fail.


maxbube:
Hi @tanji , 

makes sense :) thanks for the PR

Issue #25, opened 2016-12-22
Add an option to "purge" database before loading


enboig:
There is an option to drop table if it exists, but I would like and option to drop all tables in database before loading a backup.

Thanks

fabn:
Very useful imho, especially when loading data in development environments

enboig:
Could this be merged? I think it would be useful.
Thanks

PR #91, opened 2017-12-22
Add an option to "purge" database before loading #25

dhufnagel:
Added option to purge data in tables before restoring.


maxbube:
Hi @dhufnagel, thanks for the PR but 2 things:

1. That code will remove all data already restored if you have chunks
2. I don't think a restore tool should be deleting data at all. That can be dangerous and I will let the DBA to sanitize the database server before a restoration.

athompso:
Er, @maxbube , myloader already has the -o / --overwrite-tables switch, which deletes data.  How is this different?  This is actually slightly less invasive, as table-level permissions would be preserved.

@dhufnagel IMHO you would be better served by using MySQL's TRUNCATE TABLE command, not DELETE FROM.  Per the manual,
> Truncate operations drop and re-create the table, which is much faster than deleting rows one by one, particularly for large tables.

I've got individual tables that are ~100GB in size with millions of rows; DELETE FROM will take days to complete, whereas TRUNCATE completes within seconds.  TRUNCATE is not optimal for all scenarios, however - for example, when foreign keys are present and you want the CASCADE action to happen, DELETE is required.


I was wishing myloader had this option just today, in fact, because manually deleting data from ~99% of the tables in a database (but not all!) is a huge pain - far better if the tool can just empty out the table before re-inserting new data.  (I'm using myloader to reset a complex test environment that can't just be snapshotted.)

athompso:
Also, just nit-picking now while I'm here...
* there should be a check, disallowing `(is_schema && purge_data)`
* the option should be --delete-data, not --purge, since purge has no formal meaning in MySQL
* there should be both a --delete-data and a --truncate-tables option, per my previous comments.

(Hey, I can wish...)

davidducos:
I think that we have different options to start a restoration, the current options are:
- Do nothing: do not modify the current table
- Overwrite table: drops the table and uses the structure of the table provided
But we are mentioning 2 other options that could be useful in some cases:
- Use DELETE * FROM table_name: keeps the structure and removes the data
- Use TRUNCATE TABLE table_name: keeps the structure and removes the data

We should add just one variable like: --initial-restoration-process with this set of options: (none, drop, delete, truncate)
The default values is none and for compatibility we can keep with -o which sets drop. 
We need to include the CREATE TABLE schema files to the queue only when "drop" is used and execute the delete and truncate on each case.

davidducos:
This will be reimplemented for version 0.10.5. Any comment is welcome!

davidducos (approved):

davidducos (approved):
Tests done on branch I need to merge to start testing on master

PR #92, opened 2017-12-22
Added option to disable foreign key checks on restore

dhufnagel:


maxbube:
Hi @dhufnagel that is already being added on every sql file by mydumper 

https://github.com/maxbube/mydumper/blob/d5739cf2402a4017a291874c450e322b025a4113/mydumper.c#L2355

PR #95, opened 2017-12-28
Add Docker support

pataquets:
Add Dockerfile to enable image building. Useful for development and tests under Docker workflow.
Using the official GCC image, latest tag. More info at https://hub.docker.com/_/gcc/

Just adding files, setting working dir and running make instructions (cmake is also installed). Also, fix from #81 is needed to build.

Build:

```
$ docker build -t mydumper .
```

Run:

```
$ docker run --rm -it -v [your_host_mysqld.sock]:/var/run/mysqld/mysqld.sock] mydumper mydumper --user root [other options...]
```

FYI, there's a still quicker to test, already built image on my Docker Hub. Test it by running:

```
$ docker run --rm -it -v [your_host_mysqld.sock]:/var/run/mysqld/mysqld.sock] pataquets/mydumper-src mydumper [options...]
```

Using `--rm` instead of `-d` makes the container not go background and it to be deleted after stop. Should stop by `CTRL+C`'ing it.
You will need to add another ```-v``` switch in order to mount your host's backup dir inside the container filesystem.
You can replace the last ```mydumper``` by ```myloader``` in the command lines above, as needed.

Optional improvement to come (maybe in another issue):
- Create an 'official', based on your repo, automated build at Docker Hub for the image: https://docs.docker.com/docker-hub/builds/ . Just requires a free Docker Hub account and a following a quick 'Create automated build' process. I'll be happy to help on it, if needed.


pataquets:
Chasing head, resolving conflicts since original PR.
@maxbube : ping

pataquets:
Hi, @maxbube.
Do you have any feedback on this? I think it would be great to have an official Docker image on Docker Hub. There are dozens of third-party public Docker images out there, and some even are not auditable on how they are being built:
```
$ docker search mydumper
NAME                        DESCRIPTION                                     STARS               OFFICIAL            AUTOMATED
mysqlboy/mydumper           mydumper for mysql logcial backups              3                                       
tophat/mydumper             mydumper & myloader from https://launchpad.n…   2                                       [OK]
lorello/mydumper            MyDumper / MyLoader v 0.9.5 utility for supe…   1                                       
yurkeen/mydumper-alpine     Alpine Linux based mydumper Docker image to …   1                                       [OK]
yfix/mydumper               Yfix dockerized mydumper tool from percona      1                                       [OK]
tgbyte/mydumper             Hourly MySQL backup based on mydumper.          1                                       [OK]
[...]
```
After merging this, it's a matter of creating a Docker Hub account, a few minutes of point and click and you'll have an [Automated Build](https://docs.docker.com/docker-hub/builds/) ready to go (alas a 'daily build' feature). I can help, if needed.

davidducos:
Hi @pataquets, can you review my changes? your file was not complete to compile with latest version

glensc (changes requested):
Documentation must be added that this Dockerfile is purely for development purposes.

As for real application use, docker multi-stage builds must be used to create an image containing only the files needed to run the application.

davidducos:
@glensc you are right, this is the first step. 
I checked in docker hub with my account but I can't select [this repo](https://github.com/maxbube/mydumper) as I'm not @maxbube. What we can do is to create an [Organization on GitHub](https://docs.github.com/en/github/setting-up-and-managing-organizations-and-teams/creating-a-new-organization-from-scratch) and then [transfer](https://docs.github.com/en/github/administering-a-repository/transferring-a-repository) MyDumper to this new Organization. 
The alternative is brach after each release to my account and build it, which is not what we should be doing, but as a workaround, it might work.

pataquets:
> Hi @pataquets, can you review my changes? your file was not complete to compile with latest version

I've done a couple improvements. See commit messages for details.

pataquets:
> Documentation must be added that this Dockerfile is purely for development purposes.
> 
> As for real application use, docker multi-stage builds must be used to create an image containing only the files needed to run the application.

Agreed. Just a multi-stage build cherry-picking the compiled binaries from this image into a slimmer one will do, AFAICT.

pataquets:
Looking at the changes, now there is a dependency on Percona libs. Does this means that it will not work in mainline standard (ie non-Percona) servers? If it does not, I guess there should be a warning somewhere.
Also, I've attempted to avoid the dependency on `lsb-release` package for downloading Percona libs but, after succeeding at it, I've discovered that Percona Deb depends on it for installing.
Let me know if I can be of any further help.

davidducos:
@pataquets I'm going to move this to next release, I'm sorry that we are not going to be able to do it for this release as it was planned. I'm committing to push this for the next release.

grooverdan:
Just a quick alternate way is to use your release artifacts on top of a reasonably minimal re-distributable Red Hat image:
```
FROM registry.access.redhat.com/ubi8/ubi
ARG VERSION=0.10.5
RUN dnf --disableplugin=subscription-manager install -y https://github.com/maxbube/mydumper/releases/download/v$VERSION/mydumper-$VERSION-1.el8.x86_64.rpm
```

pataquets:
Thanks for pointing out, @grooverdan. In fact, that would be valid for the distribution use case.
But it would prevent usage for local development (hack/build/repeat). It would also lose visibility on the build process, since the build logs would be elsewhere.

grooverdan:
> Thanks for pointing out, @grooverdan. In fact, that would be valid for the distribution use case.

Isn't this the main use case? (even if the two liner is a fairly basic implementation).

> But it would prevent usage for local development (hack/build/repeat).
Right, it wasn't intended for CI.

> It would also lose visibility on the build process, since the build logs would be elsewhere.

Build steps are stored withing a container (e.g. podman inspect docker.io/library/mariadb:10.6). For build logs I expect no matter what container build process occurs it I assume will be part of the CI.

For the Dockerfile in this pr, @pataquets I attempted to run build it and it failed:
* the `gcc` base image has a  `lsb_release -sc` that returns `n/a` so its not compatible with the [percona releases](https://repo.percona.com/apt/).
* `FROM debian:oldstable` and `FROM debian:stable` with `curl` in the initial packages still fails with `dpkg: error processing package percona-release (--install):`.
* when you resolve those, recommend finishing the Dockerfile with a `FROM gcr.io/distroless/base-debian10` (https://github.com/GoogleContainerTools/distroless (or `stratch` if you do a static build)) and copying the built artifacts to make this a smaller image.

davidducos (approved):

davidducos:
image
🎉 🎉 🎉 🎉 🎉

davidducos:
I have just test this: 
```
docker run --rm -it -v /mysql/db1/mysql.socket mydumper/mydumper:v0.11.5 mydumper -h 10.0.2.15 -u david -p david -B test --stream > test.sql
```
And it worked. :smile:

fliespl:
@davidducos I think it would be cool, if newest version was always tagged by "latest".

davidducos:
Hi @fliespl, that was my first docker release, I'm not an expert on docker release procedures. I'm open to hear  any other recommendations. I will review it.

fliespl:
Understood, hence left this proposal :)

Also - I think that alpine would be better source image than ubuntu.

Curious though - why do you need perconate repos?

grooverdan:
> Also - I think that alpine would be better source image than ubuntu.

Alpine is a horrible choice when the product and its underlying libraries aren't built and tested with it. If small is the requirement then there are many ways (easiest being a multistage and/or trying to build/link statically) rather than reverting to the untested default mindset solution of alpine.

glensc:
one thing is dockerfile the totally different thing is publishing.

I recommend this from one of my projects. it builds tagged release on git tags and updates latest and additionally a build on default branch main which can be master here:
- https://github.com/Taxel/PlexTraktSync/blob/24ecc4599fcb98722a6c3b666d37d4bf1bc8beca/.github/workflows/docker-image.yml

it may be too complex due multiple arch targets and testing buy you can check older commits to see how it evolved from something simple

fliespl:
> > Also - I think that alpine would be better source image than ubuntu.
> 
> Alpine is a horrible choice when the product and its underlying libraries aren't built and tested with it. If small is the requirement then there are many ways (easiest being a multistage and/or trying to build/link statically) rather than reverting to the untested default mindset solution of alpine.

@grooverdan you are totally correct. For some reason I have thought Alpine package is also available on mydumper, which is of course not the case. Nice catch.

pataquets:
@davidducos @glensc Documentation is added on #570.

PR #145, opened 2018-07-18
Add csv output option

davistardif:
--csv-output


braco:
@davistardif: csvs can be pretty treacherous in practice due to how frequently commas appear in datasets. Fields must be quoted, but then quotes themselves are common.

How hard would it be to add in a tsv (tab) option?

davistardif:
@braco that would be fairly simple. At that point, why not just allow the user to specify a delimiter?

braco:
True, maybe the switch should become something like
--delimited-text-format
which defaults to tab or comma, with an optional
--field-delimiter <custom character>

This is how mysqldump handles it:
https://dev.mysql.com/doc/mysql-backup-excerpt/5.7/en/mysqldump-delimited-text.html

davistardif:
I'm not sure that commas and quotes inside a field will cause any issues for the csv output. We already have to enclose commas and escape quotes to generate an insert statement (because commas and quotes are part of the syntax), so the same procedure should work for csv. 

It does not output column headers.

glensc:
rather off-topic, but maybe inspiring to get this merged sooner.

a great blog post how the csv is useful and can be used: https://www.percona.com/blog/2018/02/22/restore-mysql-logical-backup-maximum-speed/

davistardif:
I'd be happy to polish up this PR if I could get some direction about what it needs to be accepted.

braco:
@maxbube why was this never merged?

davidducos:
Hi @braco , if we want to proceed on this, we need myloader support from @davistardif or anyone that want to update this PR. 
As it is, I will not be reviewing/approving as it is incomplete.

davidducos:
For myloader, we should be using `LOAD DATA LOCAL INFILE '<TABLE_NAME>' REPLACE INTO TABLE`

LOCAL to avoid the security requirements and avoid the need of moving the files
REPLACE, I found that the behavior changes

We need to decide if we are going to be using specific delimiters or if we are going to allow user to change them.

davidducos:
Hi @davistardif, 

I really appreciate your contribution, it was very helpful to see another opportunity that we had. That is why I developed this: https://github.com/maxbube/mydumper/pull/307 which is a wider solution, as you can export CSV and load data. 

I will close this in favor of #307 , I hope you understand!

PR #149, opened 2018-08-19
adding support for no-create and homebrew build support

davembg:


davidducos:
This PR has 2 changes:
- Fix the compilation on OSX: this is a bug, I'm confirming it. I'm not sure that you PR will work on all cases, so I will be testing 
- Adding --no-create option: This is an enhancement. However, I don't see why we need it. Currently we have other alternatives to avoid this in myloader with https://github.com/maxbube/mydumper/pull/91. Do you still want it? as, from my point of view, I prefer to have the create table statement and do not use it, than don't have it and need it. Please, provide a use case where you see that it is useful and I will be glad to push it.

davidducos:
Closing in favor of https://github.com/maxbube/mydumper/pull/313

PR #150, opened 2018-08-28
Update README.md

yakimov:
Correction of the reference to the last version for Ubuntu


davidducos:
There is already a new version, that is why I will close this, doesn't make sense at this point in time.

Issue #151, opened 2018-08-29
ERROR 3144 (22032): Cannot create a JSON value from a string with CHARACTER SET 'binary'


bajrang0789:
MyDumper converts the CHARSET to binary while dumping 
/*!40101 SET NAMES binary*/;
/*!40014 SET FOREIGN_KEY_CHECKS=0*/;

/*!40103 SET TIME_ZONE='+00:00' */;

The issue is only when we have a VIRTUAL GENERATED COLUMN from a json field, we seem to have ERROR while Loading through MyLOADER

 ** (myloader:27658): CRITICAL **: Error restoring test_db.tbl_data_info from file test_db.tbl_data_info.00000.sql.gz: **Cannot create a JSON value from a string with CHARACTER SET 'binary'.**

Sample Table structure: - 
```
CREATE TABLE `tbl_data_info` (
  `id` bigint(20) NOT NULL AUTO_INCREMENT,
  `some_id` bigint(20) NOT NULL,
  `number` bigint(20) NOT NULL,
  `text` json NOT NULL,
  `object_id` bigint(20) GENERATED ALWAYS AS (json_unquote(json_extract(`text`,'$.ObjID'))) VIRTUAL NOT NULL,
  `status` smallint(6) NOT NULL,
  `effective_from` date NOT NULL,
  `effective_to` date NOT NULL,
  `created_at` bigint(20) NOT NULL,
  `updated_at` timestamp(3) NOT NULL DEFAULT CURRENT_TIMESTAMP(3) ON UPDATE CURRENT_TIMESTAMP(3),
  PRIMARY KEY (`id`)
) ENGINE=InnoDB AUTO_INCREMENT=10 DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci;
```
The Source and Destination have the same CHARSET and Collations as below: 
MySQL version : 5.7.17-13-log (Source)
MySQL version : 5.7.22-22-log (Dest)

```
mysql> show variables like '%collation%' ;
+----------------------+--------------------+
| Variable_name        | Value              |
+----------------------+--------------------+
| collation_connection | utf8mb4_unicode_ci |
| collation_database   | utf8mb4_unicode_ci |
| collation_server     | utf8mb4_unicode_ci |
+----------------------+--------------------+
3 rows in set (0.01 sec)

mysql> show variables like '%chara%' ;
+--------------------------+----------------------------+
| Variable_name            | Value                      |
+--------------------------+----------------------------+
[...]

PR #154, opened 2018-09-10
change to utf8mb4

bajrang0789:
change to utf8mb4 as defaults.


bajrang0789:
@glensc 
running mydumper on the server sets the global name to binary.
There is this bug which comes into the picture:
https://bugs.mysql.com/bug.php?id=88288
Only affected when we have GENERATED COLUMNS with VIRTUAL ., (STORED COLUMNS works fine)

Unless we re-run the
`SET NAMES utf8;`
on the source server all the SELECT's and INSERT's on the table: tbl_data_info REPORTS error:
```" ERROR 3144 (22032): Cannot create a JSON value from a string with CHARACTER SET 'binary'."```

**If this can pick up the server-defaults instead of binary**, By-default MySQL dump sets default-character-set to utf8. SO this issue is not encountered in mysqldump.

I tried replacing this with utf8mb4, and the error was gone.

should try to use the server-defaults, and not hard code to `binary`, 
or else do a conversion : `utf8mb4` CHARACTER SET which is compatible for all including the JSON virtually generated COLUMNS.

**NOTE:- 
The bug has been fixed on MySQL 8.0, The affected version are still with versions <= 5.7.

**Comment by bajrang0789** [comment]
TBH I didn't find any Perf impacts changing to utf8mb4, working with 167GB of data.
Also, I don't see any such binary CHARSET being set on mysqldump as well. (It picks the mysql_universal_client_charset Defaults from client character set)
https://github.com/mysql/mysql-server/blob/5.7/client/mysqldump.c#L182-L188

Moreover, this can **break the Production Reads** if you are running a Slave on MySQL 5.7, having tables with JSON virtually generated columns (mydumper is run on it)

**Comment by davidducos** [comment]
@bajrang0789 ,
Don't you think that might be simpler, if we add a variable instead of set the value in the source? If you agree, I will be closing this and create a pull request to make it possible. Please. let me know your thoughts.

**Comment by bajrang0789** [comment]
> @bajrang0789 ,
> Don't you think that might be simpler, if we add a variable instead of set the value in the source? If you agree, I will be closing this and create a pull request to make it possible. Please. let me know your thoughts.

yes, @davidducos go ahead. 
This would be a quick fix for the past issues when dealing with json related fields(bug)

```
The ideal scenario: would be great if we can set this basis on the source system character set.
But this might take into a lot of code changes to handle this for each `thrconn`
```
Thanks,

bajrang0789:
report: #151

davidducos:
@bajrang0789, 
I will be working on this when I have some time, but take into consideration that your branch will not compile anymore. 

Cheers

davidducos (approved):
Adding old request but this is something that we should take into account for future changes

bmildren:
Hi @maxbube / @davidducos 👋  🙂 , out of interest, do we know when this one might make it into a release?  We've been hit by a few sporadic replication errors related to this, and are eager to take advantage of these changes 🙂  👍

davidducos:
Hi @bmildren, It has been merge to main, you can compile to get it. The next release will be on May.

bmildren:
cool ok, thanks for the confirmation @davidducos! 🙂

PR #168, opened 2018-11-28
mydumper: fix an "SQL injection" issue when table name contains a ' or \

kennytm:
In `detect_generated_fields()` and `get_insertable_fields()`, the query was constructed by `printf`ing the database and table names without escaping. When these names contain a `'` or `\`, the query would become invalid.

This PR fixes the issue by switching to use prepare statements, so that the SQL engine could automatically escape these strings.


davidducos:
Hi @kennytm,

With latest #399 merge, I think that this is not an issue anymore, can you confirm?

Issue #170, opened 2018-12-06
Add support for chunking by partition name


stoty:
One of the reasons for using partitioned tables is the ability to independently handle the data in the partitions.

Currently mydumper is not partition aware, and there is no way to dump (or load) specific partitions of a table.

I propose adding an option that chunks the dump files by partition name. Apart from having similar advantages to the existing chunking options, this makes it possible to pre-process the dump directory, and pare it down to the desired partitions before loading.

druud:
In addition to that, I would like to see an option that adapts to the partitioning of a pre-created restore-target. Probably needs a new issue-report.

This to create dump-files that can be loaded much faster into the new target-partitions, using exactly one child process per partition. These dump-files then need to be in perfect new-PK-order.

This is useful for migrations related to changed partitioning.

"Share anything on the (constant) reading side, share nothing on the (dynamic) writing side."

Andor:
Seems like it is almost the same feature request as this old one https://bugs.launchpad.net/mydumper/+bug/1262706

stoty:
Thanks a lot for getting the feature in!

PR #171, opened 2018-12-06
Split partition

stoty:
This adds the feature proposed in issue #170.

The feature is compatible with size-based chunking, but overrides row-based chunking on partitioned tables.

The dump data file name format is extended by an additional partition section between the table name and the part section. 

Also, the semantics of the **where** field in the dump job struct is slightly changed, it now explicitly includes the _where_ or _partition_ keywords instead of just the where expression.

The resulting dumps are readable by an unpatched myloader, the changes are only made to enable logging of the extended part section.

### DIFF
--- a/mydumper.c
+++ b/mydumper.c
@@ -571,7 +571,7 @@ void *process_queue(struct thread_data *td) {
 		switch (job->type) {
 			case JOB_DUMP:
 				tj=(struct table_job *)job->job_data;
-				g_message("Thread %d dumping data for `%s`.`%s`%s", td->thread_id, tj->database, tj->table, (tj->where == NULL) ? "" : tj->where);
+				g_message("Thread %d dumping data for `%s`.`%s`%s", td->thread_id, tj->database, tj->table, tj->where ? tj->where : "");
 				if(use_savepoints && mysql_query(thrconn, "SAVEPOINT mydumper")){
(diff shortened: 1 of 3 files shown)


stoty:
Unfortunately, I don't expect to have the time to work on this in the foreseeable future.
Thanks for taking up maintainership, though.

davidducos:
We need to reimplement this with a newer version to avoid merging so many lines. I need to go forward with other PR to be finally merge this.

davidducos:
Hi @stoty ,
I need to apologize, but I will need to postpone this another release, hoping to don't postpone again. It is not that I consider this useless, for the contrary I think that it will be very useful, but I need to work/think on the filename formatting a bit longer before changing it.

davidducos:
I will circle back to it after merging https://github.com/maxbube/mydumper/pull/414. 

@stoty, I will need access to your repo to apply all the commit that master has, and then implement the merge your code. Let me know when I can start working on it.

stoty:
Thank you @davidducos .

The https://github.com/stoty/mydumper/tree/split_partition repo is a public one, you can check it out, and perform any rebase, merge, etc that you need to do on your own branch.
You can close this PR and open one from your own branch with the rebased change.

stoty:
I've squashed the commit for easier rebase, and updated the author information.

davidducos:
This was implemented in #521

Issue #172, opened 2018-12-06
triggers restore(wrong pattern?)


tsolodov:
step on a bug:

folder with multiple DB backups.
db naming:
XXXX_YYY_1000
XXXX_YYY_1001
XXXX_YYY_1000000000000000094

When you do restore:
of XXXX_YYY_1000:
myloader -o --source-db=XXXX_YYY_1000 --verbose 3  -d ./_FOLDER_WITH_BACKUP_/  -t 3 --enable-binlog --host=MySQL_is_amazing.org

what I found:

** (myloader:13412): CRITICAL **: Error restoring XXXX_YYY_1000000000000000094.table from file XXXX_YYY_1000000000000000094.table-triggers.sql: This version of MySQL doesn't yet support 'multiple triggers with the same action time and event for one table'.

My guess: after restoring data looks like myloader trying to load all triggers with "XXXX_YYY_1000*" pattern, so XXXX_YYY_1000000000000000094 database is matched.

tsolodov:
Created & tested fix: 
https://github.com/tsolodov/mydumper/commit/ae9f08d3b64a08315e168235f017d88b6b7f7eed

maxbube:
Thanks for PR @tsolodov!

PR #173, opened 2018-12-06
Fixed issue #172

tsolodov:
Fix for issue #172 . I compiled and tested, it works for me

### DIFF
--- a/myloader.c
+++ b/myloader.c
@@ -256,7 +256,7 @@ void restore_schema_view(MYSQL *conn){
 	const gchar* filename= NULL;
 
 	while((filename= g_dir_read_name(dir))) {
-		if (!source_db || g_str_has_prefix(filename, source_db)){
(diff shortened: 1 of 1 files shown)


tsolodov:
Hello dear contributors of mydumper, could you please review and merge it to master? From my perspective it's good to have it fixed.

kvasserman:
This commit actually breaks restores of routines. The code change is looking for files with <dbname> followed by a period (dot), but routines are stored in "<dbname>-schema-post" files and so they are never picked up or executed. The result is that -schema-view, -schema-trigger and -schema-post files are ignored.

tsolodov:
> This commit actually breaks restores of routines. The code change is looking for files with followed by a period (dot), but routines are stored in "-schema-post" files and so they are never picked up or executed. The result is that -schema-view, -schema-trigger and -schema-post files are ignored.

Hey,

this bug was already addressed https://github.com/tsolodov/mydumper/commit/ae9f08d3b64a08315e168235f017d88b6b7f7eed

PR #186, opened 2019-05-02
clang-format -i -style=llvm -sort-includes=FALSE *.{c,h}

morgo:
I would like to propose that clang-format be added as a pre-commit hook. Particularly with `mydumper.c` it is hard to keep track of some of the indentation because it is such a long file.

This command just runs `clang-format -i -style=llvm -sort-includes=FALSE *.{c,h}. I tried with include sorting on (the default) and it breaks the build.

It makes some of the comments break in weird places, but that could maybe be fixed in followup work. I thought it was important to show what would change without sneaking other changes in.

### DIFF
--- a/binlog.c
+++ b/binlog.c
@@ -12,23 +12,23 @@
     You should have received a copy of the GNU General Public License
     along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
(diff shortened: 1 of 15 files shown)


maxbube:
Hey @morgo I applied clang after all merges and changes for next release. I will proceed to close this PR and thanks!!!

PR #189, opened 2019-05-06
Remove drizzle support

morgo:
This PR does the following:

* Removes support for drizzle, and simplifies a number of if statements.
* Previously MariaDB was flavor detected and then handled the same as MySQL, but this is likely to cause problems already. So I have changed it to say it is "MariaDB".

After this PR, the only special flavor detection is for TiDB, although within the code there is version detection. In future it makes sense to extend the flavor detection to also handle Percona Server because of this line in `mydumper.c`:

```
      if (!g_str_has_prefix("Percona", ver[0]) &&
```

I ran `clang-format` over files after editing them to ensure indentation was fixed.


davidducos:
@morgo there a couple of reviews in the code, let me know your thoughts, so I can merge this code.

morgo:
> @morgo there a couple of reviews in the code, let me know your thoughts, so I can merge this code.

I can't see them yet. Is it possible they are still a draft?

davidducos:
@morgo can you check again?

morgo:
> @morgo can you check again?

sorry, there are no review comments yet :(

davidducos:
I'm going to close this and create a new issue to review the code when DETECT_SERVER is being used

PR #190, opened 2019-05-10
use configurable compression commands, e.g. rsyncable gzip

Catscrash:
at the moment zlib is used to create gzip compressed files. Unfortunately this does not support rsyncability, or other compression formats.

This Pull Request adds --compress_command and --compress_extension. If --compression is set, --compress_command and --compress_extension must be set

You can then call
mydumper -h [...] --compress -compress_command "/bin/gzip --rsyncable" --compress_extension gz

unfortunately, myloader is not changed yet - for gzip no change is necessary, if someone chooses to use bzip2 or similar, myloader won't be able to load directly. This needs to be done and will come in the next few days

### DIFF
--- a/mydumper.c
+++ b/mydumper.c
@@ -26,7 +26,6 @@
 #if defined MARIADB_CLIENT_VERSION_STR && !defined MYSQL_SERVER_VERSION
 #define MYSQL_SERVER_VERSION MARIADB_CLIENT_VERSION_STR
 #endif
-
 #include <unistd.h>
 #include <stdio.h>
 #include <string.h>
@@ -78,6 +77,8 @@ int skip_tz = 0;
 int need_dummy_read = 0;
 int need_dummy_toku_read = 0;
 int compress_output = 0;
+gchar *compress_command = NULL;
+gchar *compress_extension = NULL;
 int killqueries = 0;
 int detected_server = 0;
 int lock_all_tables = 0;
@@ -168,6 +169,8 @@ static GOptionEntry entries[] = {
      NULL},
     {"compress", 'c', 0, G_OPTION_ARG_NONE, &compress_output,
      "Compress output files", NULL},
+    {"compress_command", 'C', 0, G_OPTION_ARG_STRING, &compress_command, "Compress command, mandatory if compress is active (e.g.: /bin/gzip --rsyncable", NULL},
+    {"compress_extension", 'X', 0, G_OPTION_ARG_STRING, &compress_extension, "Filename extension for compressed files, mandatory if compress is active (e.g.: gz)", NULL},
     {"build-empty-files", 'e', 0, G_OPTION_ARG_NONE, &build_empty_files,
      "Build dump files even if no data available from table", NULL},
     {"regex", 'x', 0, G_OPTION_ARG_STRING, &regexstring,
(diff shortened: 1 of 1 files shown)


mydumper.c
```diff
@@ -3699,7 +3738,7 @@ gboolean write_data(FILE *file, GString *data) {
     if (!compress_output)
       r = write(fileno(file), data->str + written, data->len);
     else
-      r = gzwrite((gzFile)file, data->str + written, data->len);
+      r = write(fileno(file), data->str + written, data->len);
```
ChristopherSchultz: Seems like the "if" and the "else" have the same code, now, yes?

mydumper.c
```diff
@@ -1998,9 +2024,13 @@ void dump_create_database_data(MYSQL *conn, char *database, char *filename) {
 
   if (!compress_output)
     outfile = g_fopen(filename, "w");
-  else
```
ChristopherSchultz: If `outfile` is now `FILE*`, then it can be redefined as such.

mydumper.c
```diff
@@ -1998,9 +2024,13 @@ void dump_create_database_data(MYSQL *conn, char *database, char *filename) {
...
-    outfile = (void *)gzopen(filename, "w");
-
+  else {
+      char *command = g_strdup_printf("%s > %s", compress_command,filename);
+        outfile = popen(command, "w");
```
ChristopherSchultz: The `popen` call can be very dangerous. Is input sanitization warranted, here, or at least a *big scary warning* about delegating to the shell?

mydumper.c
ChristopherSchultz: Can all references to `zlib` be removed, now?

Andor:
What about the same command for `myloader`?

davidducos:
If we want to continue with this, we need full support for myloader too.

davidducos:
Hi @Catscrash, 
We need to decide between 2 paths, one is to merge #403 and #336, the other option is implement what you are suggesting. Both has pros and cons, and actually #336 has a real conflict with this solution. If you consider that we must add it and you implement the same for myloader, then I will work on merging it, but until that time came I will push it back. 
Please, let me know your thoughs.

Catscrash:
@david-ducos-percona you are right, I also never got around to make the change in the loader so far.

As far as I understood, you are suggesting, that you first go ahead with the other changes and I will then check on if and how my idea still makes sense and I would then redo it? Sounds like a plan.

davidducos:
@Catscrash 
I have just realized that we can implement this but as we implemented stream, which might give us better concurrency, control and use cases. I will be creating a new pull request and close this one as the implementation will be very different. I will be linking both issues and mention you as I will be using pieces of your code.

davidducos:
Implemented in #709

Catscrash:
Hi @davidducos one thing I noticed that is not optimal yet with the exec parameter is that the original file is first being written completely to disk before it executes the script, instead of piping the data directly into the exec script. With very large tables this leads to a lot more space usage during the backup progress. This is something that was not the case in my version, do you think this is something that we could build into exec?

davidducos:
Hi @Catscrash,
You are right in how this has been implemented. Actually exec is single threaded but we could do it multi threaded in future releases. About the disk space, it will be an issue if the process that you are executing is taking longer that taking the data chunk out of the database, but you are right that we will need more disk space for buffering the data that is being processed. You can find more info about how it was implemented stream in [here](https://www.percona.com/blog/mydumper-stream-implementation/) which similar on how we implemente exec. I think that this will speed up the export stage, as in your implementation we need to wait until your process finishes to process the data to start with the next chunk... 
Both implementations has pros and cons. I think that adding more threads to --exec, might speed up the overall process but it will not solve the disk buffering situation that you want to avoid.

Catscrash:
@davidducos ok, on the one hand that makes sense... But with --compress, we don't write it to the filesystem before compressing either and it's still plenty fast. If we had a way of making it optional, wether we give the exec script the data stream without writing it first, or the current way, I think we would have best of both worlds?

davidducos:
@ChristopherSchultz are you looking #709 ??

@Catscrash, --compress is prior --exec, and you can use both. You can use --compress which will be sending to --exec a compressed file to be processed. 

So, what you want is, instead of writing to file, to write into a pipe. Am I right? (I'm still thinking how to implement... thinking...)

davidducos:
@Catscrash, ok... but I still don't like it. I see no performance gain on this solution, actually it slows down the overall process if the postprocessing task is slower than write to disk. The only reason of having an --exec-per-thread is because you don't want the disk buffer, not because of the space, I was thinking on security as it will not be encrypted. 
Anyways, it will be good to have.

PR #193, opened 2019-06-04
Fix download urls for Debian / Ubuntu

Djailla:
The download urls have changed, fix it in the README


alekseyp:
https://github.com/maxbube/mydumper/pull/174

Issue #141, opened 2018-06-05
Table checksums?


stephenreay:
I’ve tried searching and read various versions of the man page but I can’t find a Mydumper option (or even an external solution short of using a node without any write activity) to write out the results of `checksum table` once the transaction has started, to allow for easier verifying following a restore.

maxbube:
Hi @stephenreay is maybe pt-table-checksum what you need after a restore? you can find it here https://www.percona.com/doc/percona-toolkit/LATEST/pt-table-checksum.html

stephenreay:
Hi @maxbube I'm not sure how that would really work in a backup/restore scenario.

If `mydumper` could store the result of `checksum table` on each table it's dumping, *after* it has started the transaction (and store it in a file alongside the dump), either a custom tool, (or ideally `myloader`) could then use that to ensure the restored table matches the original one, even if the original is lost.

stephenreay:
Hi @maxbube,

Let me preface this by saying: I don't consider C or C++ to be things "in my wheelhouse".

Despite the above, I've managed to get this minimally working (i.e. it generates a new `db.table.checksum` file with the appropriate value in it, unless `--no-checksums` option is given).

Would you consider a PR for this feature (and forewarning, I absolutely expect some of what I've added will need attention to make it acceptable to include in upstream)?

The obvious (to me at least) next step would be support for reading the checksum values in `myloader` and throwing an error if a table doesn't match the stored checksum after loading the data. I haven't even looked at the `myloader.{c,h}` files yet, (so far, I'm quite honestly just happy that it still compiles with my changes and works), but if you'd consider merging this PR I'm happy to have a go at getting it to check them when loading too.

davidducos:
Hi @stephenreay, I found a bug and push a PR with the fix. 
I realized of the bug when I was working on the testings... looks like the testings worked in this case! 🎉 🎉

stephenreay:
Nice work.

I don't think it ever occurred to me that compression would be expected on those files, due to their fixed, small size. But I guess it makes sense for consistency.

PR #194, opened 2019-06-06
Added support for dumping checksums

stephenreay:
This is an initial change to provide the functionality requested in #141 

As mentioned in the issue - this isn't my normal gig, I'm just trying to solve a problem.


stefanhipfel:
This is something I also need for my backup scenario. This is the perfect place to get accurate checksums of tables at the moment of backup while the lock is still active!

stephenreay:
@maxbube any chance you can take a look at this soon?

@stefanhipfel haven’t found another reliable solution, sorry.

stephenreay:
Bump. @maxbube any chance you could look at this?

stephenreay:
@maxbube is there something I can do, to improve this, so it can be accepted?

davidducos:
Hi @stephenreay,

I reviewed it and if I didn't understood this incorrectly, you want the checksums to check the consistency of the backup after restore. However, I don't see any change in myloader to perform the checksum analysis. If you can add a commit with myloader support, I will be reviewing and adding suggestions. 

Thanks!

stephenreay:
@davidducos thanks for looking at this!

I didn't add loader support at the time as I was hoping for some feedback on the change to dumper (C is not my usual bread and butter language). It also seemed slightly less crucial - getting the checksum inside the transaction is essential - but in the case of a restore it's likely that no ongoing operations are running queries so it's more feasible to run a manual query after restore to compare the checksum.

However, I will endeavour to add checking support in loader. Thanks again for looking at this!

stephenreay:
@davidducos I've rebased the branch and added support for verifying checksums after loading.

It doesn't try to use the worker threads for verification, as a data load could still be in-progress on another thread. 

I'm all ears for feedback on how this might be improved.


Thanks again for looking at this!

davidducos:
@stephenreay I have several changes on my bucket but I don't have permissions to pull on your repository. Is it possible to grant my user (@davidducos) the privileges to commit on your repo/branch?

stephenreay:
@davidducos done. Sorry for the delay. Thanks!

davidducos:
@stephenreay I did a couple of changes, please, review to confirm that it is working as expected. Take into account that I change the variable name to dump_checksum and that the default is going to be FALSE by default, as it will be a negative impact in performance if it is enable. Please, let me know your thoughts.

stephenreay:
Hey @davidducos it seems to work fine still.

I swapped the help text, and I had to change the short-option, because it was conflicting with `-U, --updated-since`.
I picked `-M` just because it was the next available letter in the word checksum, but I don't think it's a big deal if it only had a long option name, either.

davidducos (approved):

PR #213, opened 2019-11-15
Fix the create database section when creating with --source-db enabled

DeLoWaN:
If you specify a --database and --source-db in the imort process, myloader statically generates a `CREATE DATABASE` instruction. But some information are missing such as CHARACTER SET or COLLATION. You end up with a different character set on you imported database.

This fix will use the content of the source-db -schema-create.sql to create the final database.

It was not an easy fix and perhaps I did it in a wrong way, but I have not done any C programming in the last 15 years !


davidducos:
I tested it and it is not doing what suppose to do and there are debugging messages that should not be there. @DeLoWaN, can you fix it?

davidducos:
Ah and by the way, this is a bug.

DeLoWaN:
Well, I retested it and it does what it supposed to do.

My system create a database with a default character set with latin1.

```
mysql> show create database source_db_utf8;
+----------------+-------------------------------------------------------------------------+
| Database       | Create Database                                                         |
+----------------+-------------------------------------------------------------------------+
| source_db_utf8 | CREATE DATABASE `source_db_utf8` /*!40100 DEFAULT CHARACTER SET utf8 */ |
+----------------+-------------------------------------------------------------------------+
1 row in set (0.00 sec)
```

Then, with `./myloader -d ~/tmp/dump/utf8 -B import_db -s source_db_utf8`

Prior my PR, the character set is set to system default:

```
mysql> show create database import_db;
+-----------+----------------------------------------------------------------------+
| Database  | Create Database                                                      |
+-----------+----------------------------------------------------------------------+
| import_db | CREATE DATABASE `import_db` /*!40100 DEFAULT CHARACTER SET latin1 */ |
+-----------+----------------------------------------------------------------------+
1 row in set (0.00 sec)
```

After my PR, the character set is the one which is exported:

```
mysql> show create database import_db;
+-----------+--------------------------------------------------------------------+
| Database  | Create Database                                                    |
+-----------+--------------------------------------------------------------------+
| import_db | CREATE DATABASE `import_db` /*!40100 DEFAULT CHARACTER SET utf8 */ |
+-----------+--------------------------------------------------------------------+
1 row in set (0.00 sec)
```

I do removed the debugging message.

davidducos:
@DeLoWaN, this is working if you set -s only:
```
root@ubuntu-focal:/tmp/pr213/mydumper# ./myloader -d  /tmp/pr213/dump/ -B import_db
mysql> show create database import_db;
+-----------+----------------------------------------------------------------------+
| Database  | Create Database                                                      |
+-----------+----------------------------------------------------------------------+
| import_db | CREATE DATABASE `import_db` /*!40100 DEFAULT CHARACTER SET latin1 */ |
+-----------+----------------------------------------------------------------------+
1 row in set (0.01 sec)
mysql> drop database import_db;
Query OK, 1 row affected (0.00 sec)
root@ubuntu-focal:/tmp/pr213/mydumper# ./myloader -d  /tmp/pr213/dump/ -B import_db -s source_db_utf8
mysql> show create database import_db;
+-----------+--------------------------------------------------------------------+
| Database  | Create Database                                                    |
+-----------+--------------------------------------------------------------------+
| import_db | CREATE DATABASE `import_db` /*!40100 DEFAULT CHARACTER SET utf8 */ |
+-----------+--------------------------------------------------------------------+
1 row in set (0.00 sec)
```
Should we use the CREATE DATABASE statement from backup always? Why are we generating our own CREATE DATABASE statement?
What do you think?

DeLoWaN:
You're right, it was not in my use case :).

I made an update to make it more general. I always took the CREATE DATABASE from dump, I realized there is no reason to try to read from the `db` variable since it's the destination database.

I don't really know why would we need the line 372:

```
query = g_strdup_printf("CREATE DATABASE `%s`", db ? db : database);
```

But it was there before, I just leave it be.

myloader.c
```diff
@@ -355,25 +355,19 @@ void create_database(MYSQL *conn, gchar *database) {
...
+  const gchar *filepathgz = g_strdup_printf("%s/%s-schema-create.sql.gz",
+                                            directory, source_db ? source_db : database);
+
+  if (g_file_test(filepath, G_FILE_TEST_EXISTS)) {
+    restore_data(conn, db ? db : database, NULL, filename, TRUE, FALSE);
+  } else if (g_file_test(filepathgz, G_FILE_TEST_EXISTS)) {
+    restore_data(conn, db ? db : database, NULL, filenamegz, TRUE, FALSE);
   } else {
     query = g_strdup_printf("CREATE DATABASE `%s`", db ? db : database);
     mysql_query(conn, query);
```
DeLoWaN, line 373: I don't see a use case where this query is really needed.

davidducos (approved):

Issue #215, opened 2019-12-03
Error dumping table with GENERATED ALWAYS column


DmitriyLyalyuev:
Hi. 
Thanks for the great tool.

We are trying to use `mydumper` to dump DB with a table which includes GENERATED ALWAYS column, but no luck. 

We are getting an error:

```
** (mydumper:23897): CRITICAL **: Error dumping table (database.core_assessment_term_grades) data: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'created_at,updated_at,core_unit_id,type FROM `database`.`core_asse' at line 1
```

mysqldump doing dump well. 

There is the table definition:

```
CREATE TABLE `core_assessment_term_grades` (
  `id` bigint(20) NOT NULL AUTO_INCREMENT,
  `user_id` int(11) NOT NULL,
  `academic_term_id` int(11) NOT NULL,
  `ib_class_id` int(11) NOT NULL,
  `values` json NOT NULL,
  `created_at` datetime NOT NULL,
  `updated_at` datetime NOT NULL,
  `core_unit_id` int(11) DEFAULT NULL,
  `core_unit_id_not_null` int(11) GENERATED ALWAYS AS (coalesce(`core_unit_id`,0)) STORED NOT NULL,
  `type` varchar(255) NOT NULL,
  PRIMARY KEY (`id`),
  UNIQUE KEY `index_core_assessment_term_grades_unique` (`ib_class_id`,`academic_term_id`,`user_id`,`core_unit_id_not_null`),
  KEY `index_core_assessment_term_grades_on_user_id` (`user_id`),
  KEY `index_core_assessment_term_grades_on_ib_class_id` (`ib_class_id`),
  KEY `index_core_assessment_term_grades_on_core_unit_id` (`core_unit_id`),
  KEY `index_core_assessment_term_grades_on_academic_term_id` (`academic_term_id`),
  CONSTRAINT `fk_rails_26fa6260b4` FOREIGN KEY (`academic_term_id`) REFERENCES `academic_terms` (`id`),
  CONSTRAINT `fk_rails_9ea8694a96` FOREIGN KEY (`user_id`) REFERENCES `users` (`id`),
  CONSTRAINT `fk_rails_af24801b3b` FOREIGN KEY (`ib_class_id`) REFERENCES `groups` (`id`)
) ENGINE=InnoDB AUTO_INCREMENT=2523015 DEFAULT CHARSET=utf8mb4;
```

[...]

PR #216, opened 2019-12-03
Add quoting to column names when generating string for SELECT

fredricj:
get_insertable_fields() is missing quoting of the column names when generating the string of column names to return.

Fixes issue #215

### DIFF
--- a/mydumper.c
+++ b/mydumper.c
@@ -2112,7 +2112,9 @@ GString *get_insertable_fields(MYSQL *conn, char *database, char *table) {
       g_string_append(field_list, ",");
     }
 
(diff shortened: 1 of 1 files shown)


fredricj:
Yeah, it did. Fixed now

Tominounet:
Hello everyone.
Just wanted to notice you that my team is using mydumper and myloader, and we had the bug for dumping table with generated columns.
So we tried to apply this fix and we test it... and it works perfectly !
You can merge this fix and the master I think, great job guys.

maxbube:
Hi! Thanks for the contribution, I already merged this PR. But I can't commit to a release date.

PR #227, opened 2020-01-24
Order by primary key

### LINKED ISSUE #48: add option to order rows by PK
Would  be useful to have mydumper implement the option similar to mysqldump.
Per mysqldump manual:
--order-by-primary

Dump each table's rows sorted by its primary key, or by its first unique index, if such an index exists. This is useful when dumping a MyISAM table to be loaded into an InnoDB table, but makes the dump operation take considerably longer.

fredricj:
Add option to order by Primary key or Unique key if no primary key exists in table
Fixes #48

### DIFF
--- a/mydumper.c
+++ b/mydumper.c
@@ -525,6 +525,13 @@ void write_snapshot_info(MYSQL *conn, FILE *file) {
     mysql_free_result(mdb);
 }
 
+void message_dumping_data(guint thread_id, struct table_job *tj){
+  g_message("Thread %d dumping data for `%s`.`%s`%s%s%s%s",
+                    thread_id, tj->database, tj->table, 
+		    tj->where ? " WHERE " : "", tj->where ? tj->where : "",
+                    tj->order_by ? " ORDER BY " : "", tj->order_by ? tj->order_by : "");
+}
+
 void *process_queue(struct thread_data *td) {
   struct configuration *conf = td->conf;
   // mysql_init is not thread safe, especially in Connector/C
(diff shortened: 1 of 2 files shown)


raags:
Will this work when using the `--rows` option as well? But I suppose that will need changes in `myloader`.

fredricj:
From just a quick look I think the dumping will work fine with --rows and --chunk-filesize. Loading is probably fine when threads=1 . With threads>1 it will likely not work as expected since it will do multiple chunks for the same table simultaneously even if other tables are pending (I think). Not sure how you can solve that without rewriting the loader to only allow one thread per table

davidducos:
I will be approving this after testing, but I think that the data will be in order anyway as the clustered index will be use for innodb tables. 
I would like to leave this code as we could use it for other use cases in the future.

davidducos:
Hi @fredricj, 

Latest commit is a fix about the message, as during my testings I found no message related to the order by, so I decided to created a function that merge all possible "data dump" message cases. Can you review it?

fredricj:
Looks fine to me

davidducos (approved):

druud:
1. With enums, be aware of special index behavior: with an index, the order is by the index-number of the option; without the index, the order is by the alphabetical order of the option strings.
2. If partial indexes are involved, like `name(24)`, then best use the `HANDLER` syntax to accurately find borders in the data.
3. Multi-value comparisons, like `(c1,c2,c3)>=(v1,v2,v3)`, are still buggy (in performance), so best to still avoid.

PR #239, opened 2020-03-25
global read lock

liuqian1990:
https://dev.mysql.com/doc/refman/8.0/en/flush.html#flush-tables-with-read-lock


davidducos (approved):
This is correct

PR #244, opened 2020-05-08
Link mydumper against libm

AdrianBunk:
This is required due to mydumper.c using ceil().

https://bugs.debian.org/956020

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -37,7 +37,7 @@ if (WITH_BINLOG)
 else (WITH_BINLOG)
   add_executable(mydumper mydumper.c server_detect.c g_unix_signal.c connection.c getPassword.c)
 endif (WITH_BINLOG)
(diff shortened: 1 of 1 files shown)


davidducos:
Hi @AdrianBunk,

Should this change implemented for all the distributions of MyDumper? or just for Debian/Ubuntu releases? Please, let me know I would like to include it in the next release which will be this month.

AdrianBunk:
> Should this change implemented for all the distributions of MyDumper? or just for Debian/Ubuntu releases?

This is unrelated to the distribution.

https://www.man7.org/linux/man-pages/man3/ceil.3.html
"Link with -lm."

The compiler may emit a call to a function in libm or instructions implementing the functionality.
This is an implementation detail of the compiler that can change at any time, the application is supposed to link with libm.

davidducos (approved):
I had tested and even in the environment that I'm using there is no error, makes sense to add this change to avoid issues for others.

PR #248, opened 2020-05-21
Fix rows max to use unsigned instead of signed max

### LINKED ISSUE #245: --rows misses data when primary key is of type unsigned bigint
When the PK used is `unsigned bigint` the max that mydumper uses is `signed` bigint:
```
2020-05-16 20:11:15 [INFO] - Thread 8 dumping data for `xxxxx`.`xxxxx` where (`id` >= xxxxxxx AND `id` < 9223372036854775819)
``` 
Which is the signed max, instead of the unsigned max `18446744073709551615`. Thi misses rows with values greater than `9223372036854775819`.

raags:
The row max should be converted into unsigned long long instead of
signed. Tables with Primary keys over signed long long will be missed.

Changed other places to use unsigned long long as well to maintain
consistency.

Fixes #245

### DIFF
--- a/mydumper.c
+++ b/mydumper.c
@@ -2169,7 +2169,7 @@ GList *get_chunks_for_table(MYSQL *conn, char *database, char *table,
     while ((row = mysql_fetch_row(indexes))) {
       if (!strcmp(row[3], "1")) {
         if (row[6])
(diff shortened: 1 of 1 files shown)


maxbube:
Thanks for the contribution @raags !

Issue #182, opened 2019-03-12
Segmentation fault (core dumped)


leshalv:
执行msql远程导出命令提示Segmentation fault (core dumped)错误
![](https://user-images.githubusercontent.com/30397655/54181766-8d8e8e00-44da-11e9-94c4-9814c7b1ebd0.png)

799983050:
你的是阿里云的RDS吗？我的是报** (mydumper:22117): WARNING **: 11:21:33.693: Executing in no-locks mode, snapshot will notbe consistent
Segmentation fault

Esysteme:
same there :

```
7:28:28	bash: line 1:     6 Segmentation fault      (core dumped) mydumper -h $RD_NODE_HOSTNAME -G -R -E -u backup -p $RD_OPTION_BACKUP_PASSWORD -v 3 -r 100000 -o /export/$BACKUP_DATE
```


mydumper 0.9.1, built against MySQL 10.1.20-MariaDB

Esysteme:
this problem happen on : 
MariaDB [(none)]> select version();
+---------------------------------------------+
| version()                                   |
+---------------------------------------------+
| 10.3.12-MariaDB-1:10.3.12+maria~stretch-log |
+---------------------------------------------+
1 row in set (0.000 sec)

+ Galera Cluster

dupondje:
Any chance to test this against the current git + have a coredump?

shijiunv:
When exporting a table with commas with mydumper, system prompt that Segmentation fault (core dumped) .

shijiunv:
When I run mydumper for more than 1 hour,  Segmentation fault.

fredricj:
> 
> 
> When exporting a table with single quotes, system prompt that Segmentation fault (core dumped) .
Could you provide a test case for this?

davidducos:
We need feedback to proceed. Closing it.
https://github.com/maxbube/mydumper/wiki/Support

PR #258, opened 2020-07-15
Fix issue #182 for tables with quotes

fredricj:
detect_generated_fields() didnt escape the table and database strings so a table name with quotes caused a syntax error and in the end a segmentation fault.


davidducos:
This has been implemented in #399

PR #260, opened 2020-07-29
Skip from the processlist the system user

dupondje:
Just 2 minor fixes to fix 2 bugs.


davidducos (approved):
The code seems to be really simple and I had tested

VMEMachuca:
@davidducos it seems like this change would be available on v.0.10.3 which is expected to be out on March

PR #261, opened 2020-07-31
Update mydumper.c to avoid MyISAM dependency in temporary tables for VIEWS

davidebelloni:
Avoid MyISAM dependency for instances where this engine is in "disabled_storage_engines" or DBaaS without it


davidducos (approved):
I can't think any case where using the default storage engine could cause an issue. I'm not sure why it was forced to use MyISAM in the first place.

PR #269, opened 2020-09-20
Adding file name to the log during restoring data for better logging

punui:
As of version 0.9.5, when we throw error during the data restoration, we neither specify the thread which caused the error nor we specify the file which produced this error. Due to this, the consumer could not build a utility to find out all the files which could not be restored and thus has no option other than fully restore the db table again. 

This become worse when the chunk size in myDumper and no. of query per transaction in myloader are same and even though he/she cannot retry those files which failed due to connection issues or other temporary glitches and thus could not be restored even when the table was in stable state.


recall704 (commented):
LGTM

davidducos (approved):
Looks good to me!

Issue #221, opened 2019-12-30
Change of column extra values in the INFORMATION_SCHEMA COLUMNS table in MySQL 8.0


huangzhiyong:
In MySQL 8.0, expressions were introduced, causing the COLUMNS table column extra to change. Using mydumper would drop some columns, which caused problems.
refer: https://dev.mysql.com/doc/refman/8.0/en/columns-table.html#extra

example:
CREATE TABLE `b` (
    ->   `id` int(11) NOT NULL,
    ->   `aAt` datetime NOT NULL,
    ->   `cAt` datetime NOT NULL DEFAULT CURRENT_TIMESTAMP,
    ->   `uAt` datetime(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6),
    ->   PRIMARY KEY (`id`)
    -> ) ENGINE=InnoDB;

select COLUMN_NAME,extra from information_schema.COLUMNS where  table_name='b';
+-------------+-------------------+
| COLUMN_NAME | EXTRA             |
+-------------+-------------------+
| aAt         |                   |
| cAt         | DEFAULT_GENERATED |
| id          |                   |
| uAt         | DEFAULT_GENERATED |
+-------------+-------------------+

huangzhiyong:
mydumper detect generated fields code:
gboolean detect_generated_fields(MYSQL *conn, char *database, char *table) {
...
  gchar *query = g_strdup_printf(
      "select COLUMN_NAME from information_schema.COLUMNS where "
      "TABLE_SCHEMA='%s' and TABLE_NAME='%s' and extra like '%%GENERATED%%'",
      database, table);
  mysql_query(conn, query);
  g_free(query);
...
}

sundysundy:
mysql 8.0          like '%%VIRTUAL GENERATED%%'

fredricj:
In mysqldump the related check only looks for "virtual generated" and "stored generated" perhaps mydumper should do the same?

davidducos:
This was fixed on 0.10.3

PR #279, opened 2021-01-05
Update mydumper.c

davidducos:
I did this based on what was reported on https://github.com/maxbube/mydumper/issues/273


fredricj:
Based on a comment in #221 I think get_insertable_fields() also needs to be updated to handle DEFAULT_GENERATED

davidducos:
Thanks @fredricj branch already updated

maxbube (approved):

PR #286, opened 2021-02-05
Adding fast index creation functionality

davidducos:
This pull request uses InnoDB fast index creation by creating secondary indexes after importing the data.
Its only applies for tables with engine=InnoDB.
The change is when a schema file is processed and it is a CREATE TABLE statement.
It splits the CREATE TABLE to remove the index creation and creates a ALTER TABLE statement. 
The code uses a temporary queue which is going to keep the job that executes the ALTER TABLE.
After adding the jobs that inserts the data, it is going to move the jobs from the temporary queue and add them to the main queue.
We add the option --innodb-optimize-keys in myloader to enable this.


hanaura26 (approved):

davidducos:
Some performance impact of using it:
```
root@ubuntu-focal:/tmp/mydumper# time ./myloader -B pr_286_test --innodb-optimize-keys -v 4 -o -d export-20210331-190724/ -t 10 -L david.log

real    0m38.087s
user    0m0.344s
sys     0m0.456s
root@ubuntu-focal:/tmp/mydumper# time ./myloader -B pr_286_test -v 4 -o -d export-20210331-190724/ -t 10 -L david_2.log

real    0m55.680s
user    0m0.298s
sys     0m0.539s
```

myloader.c
```diff
@@ -44,7 +44,9 @@
 guint commit_count = 1000;
 gchar *directory = NULL;
 gboolean overwrite_tables = FALSE;
+gboolean innodb_optimize_keys = FALSE;
```
SuperQ, line 47: Any reason not to make this on by default? I can't think of any down sides.

davidducos: Not all the time it is better to use it. If you have small tables with just a few single column indexes, then you will see that there is an increase in the times which will no be more than 10%. This functionality is very useful when you have tables with millions of rows and a few of multi column indexes.

davidducos:
I'm postponing this PR because of [this issue](https://github.com/maxbube/mydumper/issues/322)

maxbube (approved):

SuperQ:
Nice!

PR #295, opened 2021-03-17
Allow retries while checking for long-running queries

verwilst:
Because it can be annoying to have mydumper just stop when there happens to be a long running query, allow multiple checks for long queries before actually failing. Both amount of retries and retry interval is configurable. Default is 0, no retries, so behaviour doesn't change.


davidducos (approved):
I did some proper changes and now it is ready to go

Issue #290, opened 2021-03-01
mydumper does not read credentials from --defaults-file


Phillip85:
Hi,
i would like to save my credentials in a file,
but it looks like --defaults-file does not accept credentials:

```
slr@shmysql1:~/backup$ mysql --defaults-extra-file=/etc/mysql/.backup-user.cnf
Welcome to the MySQL monitor.  Commands end with ; or \g.
Your MySQL connection id is 21
Server version: 8.0.22-13 Percona Server (GPL), Release '13', Revision '6f7822f'

Copyright (c) 2009-2020 Percona LLC and/or its affiliates
Copyright (c) 2000, 2020, Oracle and/or its affiliates. All rights reserved.

Oracle is a registered trademark of Oracle Corporation and/or its
affiliates. Other names may be trademarks of their respective
owners.

Type 'help;' or '\h' for help. Type '\c' to clear the current input statement.

mysql> ^DBye
slr@shmysql1:~/backup$ sh -x ./backup.sh test
+ hostname
+ date +%F_%H
+ dirname=/mnt/backup_diskstation/shmysql1-test-2021-03-01_16Uhr
+ mkdir -p /mnt/backup_diskstation/shmysql1-test-2021-03-01_16Uhr
+ mydumper --defaults-file=/etc/mysql/.backup-user.cnf --no-locks --triggers --events --routines -v 3 --threads=6 --chunk-filesize=1024 --regex ^(?!(mysql\.|information_shema\.|sys\.)) --logfile=/mnt/backup_diskstation/shmysql1-test-2021-03-01_16Uhr/logfile.log --outputdir=/mnt/backup_diskstation/shmysql1-test-2021-03-01_16Uhr
```
Logfile contains:
`2021-03-01 16:27:23 [ERROR] - Error connecting to database: Access denied for user 'xxx'@'localhost' (using password: NO)
`
the credentials working on the normal mysql shell, but not for mydumper

davidducos:
@Phillip85 

Can you execute this command:
`mysql --defaults-file=/etc/mysql/.backup-user.cnf` ??

It is just to confirm that all the information needed is in that file as:
`--defaults-extra-file=# Read this file after the global files are read.
`
and 
`--defaults-file=#       Only read default options from the given file #.
`

Phillip85:
Hi @davidducos 

yes i can:

```
slr@shmysql1:~/backup$ mysql --defaults-file=/etc/mysql/.backup-user.cnf
[...]

PR #306, opened 2021-04-06
Update mydumper_usage.rst

Phillip85:
clarify authenticate method via defaults file


Phillip85:
see #290

davidducos (approved):
This will clarify any doubt about the usage of --defaults-file

davidducos (approved):

PR #325, opened 2021-04-27
Master and Server data stored in same medatdata file, difficult to distinguish

davidducos:
The format of the master and slave status in the "metadata" file is very similar. This can cause confusion. https://github.com/maxbube/mydumper/issues/55


ehsanulrehman (approved):

davidducos:
This will be reimplemented when we implement https://github.com/mydumper/mydumper/issues/779

PR #334, opened 2021-05-05
Option to automatically disable/enable REDO_LOG

davidducos:
Adding ALTER REDO_LOG to enable and disable it


davidducos:
This PR reduces significantly the IOPS (10x) and bandwidth (4x) used during restoration.

maxbube (approved):

PR #336, opened 2021-05-07
Adding supprort for ZSTD

davidducos:
Its only support mydumper.

TODO:
- add support for myloader
- add zst extension when it is used


davidducos:
Waiting #414 to complete this implementation

maxbube (approved):

PR #339, opened 2021-05-12
Adding better doc when -DWITH_BINLOG=ON is used

davidducos:


davidducos:
It is pointless to make this change according to #341

PR #341, opened 2021-05-13
Extending support of WITH_BINLOG to MySQL 8.0

davidducos:
We had to add the binlog.cpp which is like binlog.c but we need to compile with c++ as the files that we need from the source of MySQL 8.

Another considerations is that in 5.7 I has to add this parameter to compile:
`-DMYSQL_CFLAGS="-I/usr/src/percona-server/percona-server-5.7.33-36/include"
`
But in 8 I need to include this:
`-DMYSQL_CFLAGS="-I/usr/src/percona-server/percona-server-8.0.22-13/include -I/usr/src/percona-server/percona-server-8.0.22-13/"
`
Because I need the sql/binlog.h

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 2.6)
 project(mydumper)
-set(VERSION testing)
+set(VERSION 10.6)
 set(ARCHIVE_NAME "${CMAKE_PROJECT_NAME}-${VERSION}")
 
 #Required packages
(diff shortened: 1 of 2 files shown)


maxbube:
Hi @davidducos, binlog support was deprecated and made it optional for that reason time ago, now the idea is to fully remove it. The maintenance of the code would be really hard to make it compatible across different vendors and there is already a solution for this on each distro: `mysqlbinlog`.

PR #347, opened 2021-05-19
Adding --where support

davidducos:
We are adding the --where option which is going to be attached to any SELECT that takes data from the database. This option is recommended for partial backups.

### DIFF
--- a/mydumper.c
+++ b/mydumper.c
@@ -141,6 +141,7 @@ guint updated_since = 0;
 guint trx_consistency_only = 0;
 guint complete_insert = 0;
 gchar *set_names_str=NULL;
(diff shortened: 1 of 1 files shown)


fredricj:
I havent tested it yet. I might be nitpicking but I would prefer more distinct variable names between where and where_option cause I got confused when looking at this quickly

davidducos:
@fredricj  I'm agree with you!! but I was not sure what it is going to be the best, because I think that the best is to use `where` for this new feature, so we need to decide and change the other references to something like `this_chunk` or `chunked_where_clause`. What do you think?

maxbube (approved):

PR #349, opened 2021-05-21
Escaping quotes on detect_generated_fields as it caused seg fault

davidducos:


maxbube (approved):

taddev:
Just tested and I think there might be a few more places that need to have the string escaped.

`** (mydumper:3249647): CRITICAL **: 08:11:58.172: Error: DB: Bob's dev site - Could not execute query: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 's dev site'' at line 1
`

davidducos:
Can you provide a test case? I'm not able to reproduce your issue.

taddev:
Seems to be tied to the `--events` and `--routines` options. If I take those off my command everything else runs as expected.

Issue #348, opened 2021-05-20
Segmentation fault caused by single quote in table and db name


taddev:
I'm getting a seg fault when trying to dump a database with a single quote  in the name.  It appears to not escape the quote correctly.

`** (mydumper:2116304): CRITICAL **: 14:15:49.237: Error: DB: Bob's dev site - Could not execute query: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 's dev site'' at line 1`

Version:
```
$ mydumper --version
mydumper 0.10.5, built against MySQL 5.7.33-36
```

OS:
```
$ lsb_release -d
Description:    Ubuntu 20.04.2 LTS
```

PR #350, opened 2021-05-21
Change quotes to backticks in SQL statements

taddev:
Fixes an issue when dumping routines from databases with quotes in the name.


taddev:
Related to #348 and #349

davidducos:
This PR is no longer needed as #401 already fix it

PR #370, opened 2021-06-30
Adding config file functionality

davidducos:
This PR is adding 2 features:
1- Load the parameters from a file on group section [myloader] and [mydumper]
2- Load the set session variables from [mydumper_variables] and [myloader_variables]

### DIFF
--- a/common.c
+++ b/common.c
@@ -22,3 +22,37 @@ char * checksum_table(MYSQL *conn, char *database, char *table, int *errn){
   mysql_free_result(result);
   return r;
 }
+
+
+void load_config_file(gchar *config_file,GOptionContext *context, const gchar * group ){
+  GError *error = NULL;
+  GKeyFile *kf = g_key_file_new ();
+  if (!g_key_file_load_from_file (kf, config_file,
+                                  G_KEY_FILE_KEEP_COMMENTS, &error)) {
+    g_warning ("failed to load config file: %s", error->message);
+  }
+  gsize len=0;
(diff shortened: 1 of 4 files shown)


davidducos:
The variables set in the file will be the latest to load, which means that it will get precedence from the parameters list in the execution line.

maxbube (approved):

druud:
Generally, command line option values overrule matching config-file option values, 
unless the options are of a 'strict' nature, for example to enforce policies.
IOW, config-file option values generally re-set defaults.
See also https://dev.mysql.com/doc/refman/8.0/en/option-file-options.html

For group names, maybe support a layered structure, such that [myloader_foo] is a specialization of the [myloader] settings. This also assumes that one can specify the group name suffix in the command.

davidducos:
I will check how hard it is going to be to change the variable loading order, tbh I didn't think that it would matter. 
About the %_variables, they are in a different section to make simpler the code, as it goes through the list of variables and executes them. Actually I was thinking of adding another group like mydumper/myloader_global_variables, so you can set global variables, such as innodb_flush_log_at_trx_commit, sync_binlog, and others that improves the performance and remove them as parameters. This will reduce the lines in the code and give you more options to speed up the import/export tasks. 

Take into account that we use the config file, as other tools like Proxysql or Pxc scheduler handler, which keep different groups depending what your configuring. What we are doing differently, is that we allow to keep the config for both tools: mydumper and myloader.

Please let me know if this clarifies the goal and it does or does not make sense to you

Issue #373, opened 2021-07-12
improve estimated_step


huangzhiyong:
func **get_chunks_for_table**  is used to calculate the size of each chunk. In some special cases, the value of **estimated_step** (`estimated_step = (nmax - nmin) / estimated_chunks + 1`) will be very large. It is recommended to add a global parameter. When **estimated_step** is greater than the global parameter, the block size will use the global parameter.

davidducos:
Hi @huangzhiyong , 100% agree with you!! I saw this in the past but didn't find the time to work on a solution. I really appreciate if you can add the code needed, if not, don't worry I will do it for next release. BTW, I would like to use  estimated_step as the name of the global variable and change to: `local_estimated_step = (nmax - nmin) / estimated_chunks + 1)` for the function calculation.

huangzhiyong:
Hi @davidducos, C is not my strong point, I tried to push a RP #376

PR #376, opened 2021-07-13
improve estimated_step

huangzhiyong:
limit the size of each block after the table is estimated. see #373

### DIFF
--- a/mydumper.c
+++ b/mydumper.c
@@ -142,7 +142,7 @@ guint trx_consistency_only = 0;
 guint complete_insert = 0;
 gchar *set_names_str=NULL;
 gchar *where_option=NULL;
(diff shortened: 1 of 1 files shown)


davidducos:
Hi @huangzhiyong,
It looks ok, but I think that the parameter name is confusing. I think that will be better: max-rows and clarify the difference between --rows and --max-rows, as --rows is the intention and the value that it is going to be used for estimate and --max-rows is a hard limit. What do you think?

davidducos:
Hi @huangzhiyong, please, let me know if you are not going to implement the changes requested, so I can do it myself.
Thanks!

davidducos (approved):

PR #392, opened 2021-08-06
Avoid logging SQL data on error

mgarstecki:
The deleted code logs the contents of the SQL INSERT request when the mysql query failed.

Since this content could come from any table, it can both generate high amounts of logs and potentially leak sensitive content, so I think this should not be logged at all :slightly_smiling_face:

### DIFF
--- a/myloader.c
+++ b/myloader.c
@@ -901,7 +901,7 @@ void restore_data_in_gstring_from_file(MYSQL *conn, char *database, char *table,
 {
                 if (mysql_real_query(conn, data->str, data->len)) {
 
-                        g_critical("Error restoring %s.%s from file %s: %s \n%s", db ? db : database, table, filename, mysql_error(conn),data->str);
+                        g_critical("Error restoring %s.%s from file %s: %s \n", db ? db : database, table, filename, mysql_error(conn));
                         errors++;
                         return;
                 }


davidducos:
Hi @mgarstecki, 

I'm partially agree with you. It is bad to log sensitive content, but when data is being imported, in each file could be several INSERT statements, you need to know which one failed. I think that we need to find a middle point, where we can show the user which INSERT failed without leaking sensitive content. 

Does it make sense to you?

davidducos:
Oh! I reviewed the code and I realized that it will be harder/confusing  if you have --rows. It might be just simpler to print the line where it could have broken. It might not be so simple...

mgarstecki:
The need to have accurate data to debug does makes sense to me, but the file info seems enough to me. Since large tables are chunked in multiple files, it is already easy to target the offending file for debugging :slightly_smiling_face:

mgarstecki:
> Oh! I reviewed the code and I realized that it will be harder/confusing if you have --rows. It might be just simpler to print the line where it could have broken. It might not be so simple...

Maybe this logging should be behind an explicit flag @davidducos, so the user can opt-in to having those logs knowingly ?

As it is it's quite sneaky, making it look really dangerous to me : this looks like the only place in the code (from a cursory look) where table data might be logged, and it will only happen on error, which the user might not encounter while setting up his restore pipeline.

This is exactly what happened to me : I've set up an automated restore test with `myloader`. It's cronned regularly, and has been running smoothly for a while. We've only updated recently to a version that includes this logging, and it took a few weeks to encounter an error on restore, and see this logging :grimacing:

davidducos (approved):

davidducos:
You are right, we need to remove it and add it in future iterations. 

Thanks!!!

mgarstecki:
Thank _you_ @davidducos :handshake: :slightly_smiling_face: 

As said, I think it could go behind a flag, and/or be dumped to an output file instead of logging. This would avoid the high volume of this output in the main logs and avoid data leaks, while still being able to pinpoint SQL that caused errors.
Maybe something like:
```
--sql-errors-logfile ... ["If defined, dump SQL clauses that generated errors to the associated file. Disabled by default"]
```

PR #396, opened 2021-08-15
needs to notify main thread to go ahead when "--tables-list" is used without "-B"!

qizidog:
needs to notify main thread to go ahead when "--tables-list" is used without "-B"!
otherwise main thread is blocked and just wait.

### DIFF
--- a/mydumper.c
+++ b/mydumper.c
@@ -2861,6 +2861,7 @@ void get_tables(MYSQL *conn, struct configuration *conf) {
       }
     }
   }
+  g_async_queue_pop(conf.ready_database_dump);  // needs to notify main thread to go ahead when "--tables-list" is used without "-B"!
   g_free(query);
 }
 


qizidog:
checksum function is not available for non_innodb_table when `--less-locking` is used.

davidducos:
Hi @qizidog, 
Yes! seems that there is bug, but this PR is not completely fixing it. I'm testing with:
```
rm -rf data; ./mydumper -c -o data -v 3 -T test.perftest
```
And it hangs. Please, let me know if you have more insight about it or code to push.

qizidog:
I’m so sorry, I wasn't careful enough when I submitted the code. Instead of copying the code directly from the IDE, I submitted the code directly on the web page. 

The codes I want to submit should be: 

```c
g_async_queue_push(conf->ready_database_dump, GINT_TO_POINTER(1));
```

I copied codes from my IDE directly this time. 

For some reasons, the computer I am currently using does not have a well configured development environment. Debugging and verification need to wait a short period of time (wait for the environment to be configured and get familiar with the code).

I’ll check it carefully latter!!! Sorry for the inconvenience!

davidducos:
Hi @qizidog, 
Yes, I saw that yesterday and I'm working in the solution. 
table is the the tablename at db level and table_filename is the tablename at filesystem level. Why can't they both be the same? several reasons, the most importante is that sometimes we have tablenames at db level that uses '.' (dot) which is the char that we use to split the filename in myloader to know what is the database and table name.

davidducos:
Closing it as it was already implemented in another branch. Thanks for your understanding @qizidog

PR #399, opened 2021-08-24
Table and database names changed to fix several issues

davidducos:
This PR fixes several table name related issues, but it changes the filenames. 

The behaviour of -B didn't change which means that database name will not be the same if it has been changed by mydumper.

### DIFF
--- a/mydumper.c
+++ b/mydumper.c
@@ -70,7 +70,7 @@ const char DAEMON_BINLOGS[] = "binlogs";
 #endif
 
 static GMutex *init_mutex = NULL;
-
+static GMutex *ref_table_mutex = NULL;
 /* Program options */
 gchar *output_directory = NULL;
 guint statement_size = 1000000;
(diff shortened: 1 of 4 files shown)


davidducos:
I thought that this was going to be easy. It was not. I want to include this changes on v0.10.9 as we need to fix all those issues. 
However, it means that we need to change default naming on mydumper and extra logic on myloader.
The most important changes that I need to do are:
- Any database and table name with '.' (dot) will be renamed to 'mydumper_<number>'. 
- Tables with no part will include the '00000' on the single datafile. 
Why?
We need to establish a single datafile format which is going to be: <database>.<table>.<part>.sql[.gz] and database/table can't include '.'(dots) as it will break the format.
The database and table names will be read from the -schema files.

This means that we might be facing incompatibility issues between previous backups that rely on the filename to determine the database or table name. I don't like it, but I hope to fix/allow more backups.

maxbube:
Hi @david-ducos-percona as this is breaking backward compatibility please be sure is released as 0.11.x

maxbube (approved):

PR #414, opened 2021-09-28
Using generic functions to build filenames

davidducos:
This is needed to simplify the different compression options that we want to implement


davidducos:
There will be a slight change in checksum and metadata per table files. Now it will be "-checksum"/"-metadata" instead of ".checksum"/".metadata".

PR #461, opened 2021-10-29
Adding ignore generated fields parameter

davidducos:


maxbube (changes requested):
I would change the name to no_check_generated_fields and send a warning when is set to true

maxbube (approved):

PR #465, opened 2021-11-01
Refactoring of regex to add negative regex

davidducos:


maxbube:
From my point of view, a regex is a regex and you can describe whatever you need on it. I don't think mydumper needs yet another way to filter on schemas/tables.

64kramsystem:
> From my point of view, a regex is a regex and you can describe whatever you need on it. I don't think mydumper needs yet another way to filter on schemas/tables.

I think this as a UX problem more than a technical one; while regexes can certaintly represent negative matches, the general negative match regex is very awkward. I myself use regex a lot, but I'd definitely scratch my head if I ever read one without a (explanatory) context.

For me this outweights having an extra filter (mysqlpump has actually 4+ filters :flushed:); on the other hand though, the regex for not matching a prefix is easy enough, which I guess, is the main use case, so it's not a UX tragedy :smile:.

PR #473, opened 2021-11-03
Adding IF NOT EXISTS to temporary create table for views

davidducos:


maxbube:
Still not sure about this, what if already exists with a diff definition? How can you trust the restore?

PR #479, opened 2021-11-08
[ Refactoring ] using just default config file, removing --config-file

davidducos:


maxbube (changes requested):
This is a breaking change, I guess we must first support both and print a deprecation message to not break actual deployments

davidducos:
@maxbube --config was added on v0.11.1 and as we are still on v0, I think that we can avoid formal deprecation messages, as we are planning to implement since release v1. What do you think?

maxbube (dismissed):

maxbube (approved):

PR #488, opened 2021-11-19
Setting compress_extension in the right places

davidducos:
compress_extension was not being use in all possible cases


davidducos:
Another PR fixed it.

PR #516, opened 2021-12-07
[Fix] Dynamic string length on variable

davidducos:


hubiongithub:
Hello
I tried the new version 0.11.3-4, but something seems odd, -4 is way smaller:
941744 Dez  7 11:35 mydumper_0.11.3-3.focal_amd64.deb
192372 Dez  7 15:05 mydumper_0.11.3-4.focal_amd64.deb

And:
myloader -V
myloader: error while loading shared libraries: libperconaserverclient.so.20: cannot open shared object file: No such file or directory
Perhaps the client lib was part of the .deb and is missing in the 0.11.3-4 build?

davidducos:
Hi @hubiongithub fixing the release now, the new files will be available soon.

davidducos:
Hi @hubiongithub, now you will be able to find the right files.

PR #518, opened 2021-12-07
Relocating database creation

davidducos:


davidducos:
This is just to record the solution as Master at this point, was already fixed.

PR #596, opened 2022-02-26
set verbosity level earlier

thadraibley:
The verbosity flag is initialized after attempts to read the
defaults-file argument so warnings are printed for missing
[mydumper] headers in the client file even when verbosity is
set to 1.

### DIFF
--- a/mydumper.c
+++ b/mydumper.c
@@ -1227,14 +1227,12 @@ int main(int argc, char *argv[]) {
     }
   }
 
-  set_verbose(verbose?verbose:2);
+  set_verbose(verbose);
 
   if (defaults_file != NULL){
     load_config_file(defaults_file, context, "mydumper");
   }
   g_option_context_free(context);
-
-  set_verbose(verbose);
 
   if (!compress_output) {
     m_open=&g_fopen;
--- a/mydumper.c
+++ b/mydumper.c
@@ -1227,12 +1227,14 @@ int main(int argc, char *argv[]) {
     }
   }
 
-  set_verbose(verbose);
+  set_verbose(2);
 
   if (defaults_file != NULL){
     load_config_file(defaults_file, context, "mydumper");
   }
   g_option_context_free(context);
+
+  set_verbose(verbose);
 
   if (!compress_output) {
     m_open=&g_fopen;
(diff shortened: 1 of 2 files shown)


davidducos:
Hi @thadraibley I made some changes, I basically set the verbose to 2 and then, I set to what the user has defined, does it make sense to you?

thadraibley:
Hi @davidducos, thanks for looking at it so quickly.  I understand your changes like this: If the logger was only configured before the defaults-file then any verbose options set in there would get ignored and it could only be configured as a cli switch.  But by configuring it before the defaults file as a 2 we don't get those warning messages for a missing [mydumper] section if one is not present but users that do set verbose there still get their option read and configured.  Makes sense to me, nice catch.

thadraibley:
Thinking of it though, wouldn't verbose need to be set to 1 before defaults-file is read or the same issue will occur?

thadraibley:
Yeah, on a test build this part 

`  set_verbose(verbose?verbose:2);

  if (defaults_file != NULL){
    load_config_file(defaults_file, context, "mydumper");
  }`

would have to be hardcoded to a 1
` set_verbose(1);`

since the default logging level is 2 which picks up warnings then reads the defaults-file if one was passed.

davidducos:
Hi @thadraibley, I was totally wrong from the begining! Sorry for all the noise! I'm going to revert my changes.

davidducos (approved):

thadraibley:
@davidducos  Apologies I didn't test more end cases when I submitted the PR.  Testing the changes with a defaults-file it actually looks like these changes would cause a new problem in that any settings in a defaults-file under [mydumper] would be ignored which probably isn't what would be expected, even if verbose isn't passed on the cli.  I would expect the defaults-file settings and cli options to get merged and in the case of duplicate options like verbose, one take precedence over the other but it looks like calling `set_verbose(verbose)` before `  load_config_file(defaults_file, context, "mydumper");` results in any verbosity settings inside the defaults-file to get ignored if no options are passed on the cli.  So the verbosity would always be the project default of 2, or what a user passed via the cli.  I think the ideal behavior would be: 

1. warning logs never print if verbosity is set 1 or above
2. defaults file [mydumper] settings are respected if present
3. cli passed switches are respected if passed

What do you think?

Issue #484, opened 2021-11-11
Splitting main files


davidducos:
Currently we have a thread that it is started to take a backup inside mydumper. I think that the main reason of having it is because the daemon functionality, which is ok, I'm not complaining about it. However, I would like to move the whole backup thread to another file. The main idea is to split the file as currently is large and I think that will be easier to maintain. I think that it will be useful also, if we could move the threads to a different file, at the end we might end up with this list of files:
- mydumper.c
- mydumper_main.c
- mydumper_thread.c
With myloader will be simpler, we don't need the myloader.c, so it will be:
- myloader_main.c
- myloader_thread.c

davidducos:
I also started to create files under /src. The idea will be to split functionality to make easier the review of the code, as I did in https://github.com/mydumper/mydumper/pull/549

davidducos:
I'm not going to be able to finish this for this release.

davidducos:
Mostly of the work over myloader is done. Based on the experience gained, we are going to start on mydumper for the next release

davidducos:
We were able to split mydumper.c to several files based on main functionality. I don't think that we will need to split it even more and I found really useful as I understand better the different components of the dump mechanism.

PR #605, opened 2022-03-09
Adding compression capabilities on stream

davidducos:
I had to rewrite some pieces of the stream thread as it was design to work with text files, not compressed files.


davidducos:
#484

PR #609, opened 2022-03-14
Spliting mydumper.c

davidducos:


davidducos:
Merging this. I might need to continue but I think that we can continue in the future

PR #619, opened 2022-03-21
Reengineering the DDL locks and connection sync

davidducos:


davidducos:
I think that what I have done so far is enough for this release. We need to still work in it, this is an example: https://github.com/mydumper/mydumper/issues/648.

PR #639, opened 2022-04-05
Fixing version number

ultrasilicon:


davidducos (approved):

davidducos:
Only before release we need to change to new version number, during developing they should be with even numbers

PR #697, opened 2022-05-20
Adding AFTER_IMPORT_PER_TABLE and AFTER_IMPORT_ALL_TABLES and finxing…

davidducos:
… free issues


davidducos:
On stream AFTER_IMPORT_ALL_TABLES is the default, so we are are adding the jobs to the post_queue when directory is used.

PR #500, opened 2021-11-30
More circleci building

SuperQ:
* Add more dependencies.
* Add build badge.

Signed-off-by: SuperQ <superq@gmail.com>

PR #316, opened 2021-04-22
Add minimal CircleCI config

SuperQ:
Add a simple build pipeline using CircleCI. This is just a first step.

This can be extended to test multiple combinations of different base
iamges, MySQL client libraries, and then integration tests with data.

Signed-off-by: SuperQ <superq@gmail.com>

PR #699, opened 2022-05-23
Adding checksum for table structure and procedures

davidducos:


davidducos:
With this merge, we are going to be able to perform better testings as, we are not going to use Circleci just to check if it compiles, we are going to be able to perform the test that I'm manually running.
#316 #500

PR #740, opened 2022-06-21
Fixing csv --rows issues and refactoring

davidducos:


davidducos:
Adding --load-data and --csv tests

PR #768, opened 2022-07-19
Backward compatibility for dump filenames

Fr13dr1chLuc45:
Old file names can also be exported without [.part].
This patch therefore implements a query to be able to import old exports as well.


davidducos (approved):

davidducos:
Just to clarify, this will allow you to import old files but you might have issues if the schema or table name has '.' (dots)

PR #793, opened 2022-08-14
Adding workaround to be able to share the stream temporary directory

davidducos:


davidducos:
I used this example: 
```
rm -rf data; ./mydumper -B myd_test -o data  --no-locks --stream --no-delete | grep '^\-\-' | sed 's/^/\n/g' | ./myloader --stream -d data -B myd_test_new -v 4 -o
```
This PR is not complete, as we need to add an option to avoid the usage of `grep | sed` with a single option in mydumper. I'm thinking to use --stream=[TRADITIONAL|NO_STREAM_AND_NO_DELETE|NO_DELETE], to allow multiple behaviors in the future, as I don't like the --no-delete option and I would like to remove it as it only makes sense when stream is used.

PR #803, opened 2022-08-19
Harmonised integer types for table rows to guint64

y-trudeau:
When working with large tables, I often found negative values in the table metadata files, it appears it was for table above 2^31 rows.  I harmonised the integer type to guint64.  Be aware I ran no tests (neither compiled) this change as it seems to be quite trivial.


davidducos:
Circleci ran the tests and say: [FAIL!](https://app.circleci.com/pipelines/github/mydumper/mydumper/440/workflows/d4cbba7e-15e2-4a37-bdbb-9d5bc9830e40/jobs/437)
I will check it and make the proper changes.

davidducos (approved):

PR #914, opened 2022-11-23
Build fixes

LinuxJedi:
Fixes a whole bunch of things regarding the build process, including:

* CMake deprecation warning fixed
* Human readability improvements for certain error conditions
* GLIB2 package search warning removed
* Compatibility with MariaDB Connector/C
* Fixed `make dist` to work with git instead of bzr

OpenSSL check has been removed,  mydumper itself doesn't use OpenSSL, libmysqlclient does. Checking to  see if OpenSSL is installed on the machine compiling mydumper does not  check if libmysqlclient was built with OpenSSL. libmysqlclient could for  example have been build with OpenSSL statically compiled. Also MySQL 8.0  doesn't have this as an option and MariaDB Connector/C can compile with  other SSL libraries.
    
If the intention is to check of libmysqlclient < 8.0 was built with OpenSSL then a symbol or build test check should be made instead.

Also fixes Andrew's email address / affiliation.


davidducos:
@LinuxJedi can you fix the CircleCI checks first?

LinuxJedi:
CI seems broken here, I'll closed and reopen

Issue #835, opened 2022-09-21
Use circleci to build releases


davidducos:
I'm reviewing https://circleci.com/blog/publishing-to-github-releases-via-circleci/ and I think that in the future we could start using it to draft the releases, which is going to allow us to get multiple versions, even arm64 packages.

_Originally posted by @davidducos in https://github.com/mydumper/mydumper/issues/772#issuecomment-1199693847_

davidducos:
We have an issue, after reviewing several things, I realize that CircleCI has ubuntu images only, no redhat/alma/rocky/centos images... I'm going to stop working on this until we have another alternative or option.

davidducos:
@SuperQ do you have any comments about it?

LinuxJedi:
Does the CircleCI packagecloud integration help here?

SuperQ:
@davidducos CircleCI can use any Docker image you provide. So it should be easy enough to build binaries for multiple distros.

SuperQ:
Another option would be to use GitHub Actions to build and publish things.

davidducos:
Hi @LinuxJedi / @SuperQ , thanks for your insight!
I think that I shout too early... I'm going to work a bit more on this and let you know the outcome

davidducos:
Ok, I was able to compile with Alma Linux 8. 
Now, what I'm going to be doing is build the packages and test the release creation. 
The last step will be adding the remaining OS.

BTW, I was thinking that if this works, anyone could clone this repo, made their own changes and build their own packages. That will be useful if they need to made specific changes, build and install in their env.

PR #917, opened 2022-11-24
Add ASAN and TSAN sanitizers

LinuxJedi:
New options for ASAN and TSAN sanitizers, it is recommended these are added to the CI process when issues they find are fixed.

ASAN finds leaks, similar to valgrind but is compiled in and much faster. TSAN finds thread issues such as race conditions.

You can enable them using -DWITH_ASAN or -DWITH_TSAN. They cannot be used at the same time.


davidducos (approved):

davidducos:
Hi @LinuxJedi, This failed for Bionic, Xenial and EL7.

LinuxJedi:
@davidducos  you mean when ASAN or TSAN is enabled? Yes, there are issues that need resolving. This is just the tools to detect them.

If you mean this breaks the build, do you have a link or more info? It worked on those platforms on my servers.

davidducos:
@LinuxJedi 
I rebuild my xenial VM and I got the same issue, reviewing just a bit about it I found that add_link_options was introduced in newer version of CMake that xenial doesn't have. So, in order to support old version too, I will create a PR. 
About the building process, I prefer to start working over CircleCI instead of sharing mine (https://github.com/mydumper/mydumper/issues/835) as I think that it will be more valuable.

LinuxJedi:
Ah! My bad, I already had an update CMake on my machines due to another project I work on.

davidducos:
No worries! make sense now! I will be adding the PR later today.

PR #929, opened 2022-12-01
Add `--debug` mode

LinuxJedi:
This patch does a few things:

* Add `--debug` to enable debugging output on STDOUT, this also sets `--verbose` to `3` automatically.
* Remove duplicate bit setting in `set_verbose.c`.
* Disable debugging output for the log writer if `--debug` was not specified.


LinuxJedi:
@davidducos Circle CI appears to have failed due to a network issue. I don't appear to have access to restart the run. Are you able to do it?

davidducos:
@LinuxJedi It has failed again... :(

LinuxJedi:
> @LinuxJedi It has failed again... :(

This is the good kind of failure, the one where I neglected to check that an API I used was only a year old.

davidducos:
@LinuxJedi, yeah... it happens... I will be adding testings on CircleCI for all OS that we do releases to found this kind of errors. For sure, I will be adding more testings to avoid to release with bugs...

LinuxJedi:
Fixed this by having this option only show up if you are using glib >= 2.72. Before this you'll need to use the environment variable `G_MESSAGES_DEBUG=all` (same as before this patch).

davidducos:
Thanks @LinuxJedi, I will be merging this after we fix some issues on v0.13.1 series. For sure, this will be merged by January.

davidducos (dismissed):

PR #952, opened 2022-12-21
Minor fixes to messages

LinuxJedi:
Using `--version` now shows whether it is MariaDB or MySQL that we are build with. Verbose messages now consistently use `:`.


LinuxJedi:
Closing, did it against a too-old version.

PR #962, opened 2023-01-05
Better json support

davidducos:


davidducos:
I think that, we will need to keep a sql file per db vendor, which is bad, bad, bad, because we will need to support all cases per vendor and maybe versions too.

davidducos:
I'm going to close this as there is no need to keep it, as we are going to do it in a wider project.

PR #981, opened 2023-01-16
Fixing spaces on LOAD DATA filename when compression is used

davidducos:


davidducos:
We also implemented the retry transaction in on it.

davidducos:
Hi @LinuxJedi, 

The tests against MariaDB 10.11 are failing for some bugs that we need to sort it out. As you can see from CircleCI, we are compiling for several distributions and some vendors (+ versions). Ultimate goal is to add the remaining MariaDB versions, but I would like to fix the issues first on 10.11, before going to 10.4, 10.5 and 10.6. 
What do you think?

LinuxJedi:
> Hi @LinuxJedi,
> 
> The tests against MariaDB 10.11 are failing for some bugs that we need to sort it out. As you can see from CircleCI, we are compiling for several distributions and some vendors (+ versions). Ultimate goal is to add the remaining MariaDB versions, but I would like to fix the issues first on 10.11, before going to 10.4, 10.5 and 10.6. What do you think?

Agreed, given that 10.11 is the next LTS, that should be priority. Then 10.6 and go from there. I suspect once one is fixed the rest should be quite easy.

If there isn't one already please feel free to create tickets for these issues and assign to me.

PR #988, opened 2023-01-19
Do not set UNIQUE_CHECKS=0 for MariaDB

### LINKED ISSUE #987: [BUG] myloader performance regression for MariaDB 10.6
**Describe the bug**
MariaDB 10.6 tries to accelerate loading for InnoDB tables when using `SET foreign_key_checks=0, unique_checks=0;`. This is outlined in [MDEV-515](https://jira.mariadb.org/browse/MDEV-515). Unfortunately whilst this increases single-threaded bulk insert performance this introduces table-level locks which reduce the performance of myloader.

To fix this we need to not do `unique_checks=0` for myloader for MariaDB 10.6+ restorations.

LinuxJedi:
`UNIQUE_CHECKS=0` sets MariaDB into a special mode that adds table level
locks for InnoDB. This accelerates single-threaded insert performance,
but makes multi-threaded into one table much slower.

This also adds a patch to fix the MariaDB type and version check.

Fixes #987

### DIFF
--- a/src/common.c
+++ b/src/common.c
@@ -41,7 +41,7 @@ extern int detected_server;
 
 GHashTable * initialize_hash_of_session_variables(){
   GHashTable * set_session_hash=g_hash_table_new ( g_str_hash, g_str_equal );
-  if (detected_server == SERVER_TYPE_MYSQL){
+  if (detected_server == SERVER_TYPE_MYSQL || detected_server == SERVER_TYPE_MARIADB){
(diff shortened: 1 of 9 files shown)


LinuxJedi:
Looking into CI failures...

LinuxJedi:
Should be good now, I missed a bit when cherry-picking the MariaDB ident stuff from another branch :)

PR #995, opened 2023-01-23
Adding /etc/mydumper.cnf as default file

davidducos:
This will allow use to set at config level specific configurations for different vendors and versions #987, adding particular sections, for instance:
```
[myloader_session_variables_mariadb_10_6]
SQL_MODE='NO_AUTO_VALUE_ON_ZERO' /*!40101
UNIQUE_CHECKS=1 /*!40114
FOREIGN_KEY_CHECKS=0 /*!40114
```
We are appending/replacing variables on this order of sections:
1.  `myloader_session_variables`
2. `myloader_session_variables_[VENDOR]`
3. `myloader_session_variables_[VENDOR]_[MAJOR]`
4. `myloader_session_variables_[VENDOR]_[MAJOR]_[SECONDARY]`
5. `myloader_session_variables_[VENDOR]_[MAJOR]_[SECONDARY]_[REVISION]`

The vendors supported so far are: mariadb, percona and mysql. As I'm still working on the test on tidb, it was not added yet, but we might do it in the near future.

### DIFF
--- a/mydumper.cnf
+++ b/mydumper.cnf
@@ -1,0 +1,30 @@
+[mydumper_session_variables]
+# wait_timeout = 300
+# sql_mode = 'ONLY_FULL_GROUP_BY,STRICT_TRANS_TABLES,ERROR_FOR_DIVISION_BY_ZERO,NO_ENGINE_SUBSTITUTION'
+
+[mydumper_global_variables]
+#sync_binlog = 0
+#slow_query_log = OFF
+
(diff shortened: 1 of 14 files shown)


LinuxJedi:
This minor issue here is that MariaDB currently returns that it is MySQL due to the return of the REGEX. I fixed this in #988, but feel free to fix this in another PR if needed.

PR #1145, opened 2023-05-29
Removing GZIP and ZSTD to only use --exec-per-thread

davidducos:


davidducos:
Hi @aqueos 
```
  -c, --compress                   Compress output files using: /usr/bin/gzip and /usr/bin/zstd. Options: GZIP and ZSTD. Default: GZIP
```

PR #1162, opened 2023-06-15
report error when lost mysql connectio

flttgjq:
Fix a bug may cause data loss when myloader lost mysql connection
**[descriptions]**
When there was a disconnection problem during the process of restoring data, myloader may lost data without any errors detected.
**[how did this problem happened]**
When a transaction contains lots of sql, a disconnection problem will result in data loss, think about the following scenario in myloader：
1. start transaction
2. insert_sql_1
3. insert_sql_2 (lost connection and reconnect)
4. insert_sql_3 
5. commit

We have disconnection problem while execute insert_sql_2, and we reconnect successfully, now the transaction we start before has gone, insert_sql_2 and insert_sql_3 will be autocommit because of the new connection was created. But we lost  data in insert_sql_1!
During this process, myloader did not find any errors because we reconnect and re-execute insert_sql_2 success

### DIFF
--- a/src/connection.c
+++ b/src/connection.c
@@ -86,7 +86,6 @@ GOptionGroup * load_connection_entries(GOptionContext *context){
   return connection_group;
 }
 
-gboolean reconnect = 1;
 
 void configure_connection(MYSQL *conn) {
   if (connection_defaults_file != NULL)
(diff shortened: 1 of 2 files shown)


davidducos (approved):

davidducos:
Hi @flttgjq 
Nice catch!! Thank you for finding it out!

PR #1189, opened 2023-06-29
Adding --protocol

davidducos:


davidducos:
This is the default message if you don't set anything:
```
2023-06-29 20:18:58 [INFO] - Connection via default library settings
```
If you set onnly the password:
```
2023-06-29 20:13:02 [INFO] - Connection via default library settings using password
```
If you set other details are shown like this:
```
2023-06-29 20:12:48 [INFO] - Connection via default library settings using password:
	Host: localhost
	User: root
```

Issue #1174, opened 2023-06-25
[Enhancement] Send known info that will be in metadata when --stream is used


davidducos:
When I was reviewing/implementing https://github.com/mydumper/mydumper/issues/1163#issue-1758908560, I realized that if stream is used, it was pointless. So, we need a way to send metadata information through the stream. It might be every a couple of seconds or every time that a new table name is added.

PR #1190, opened 2023-06-29
Increasing the amount of tests

davidducos:


davidducos:
Failures on the test are related to this scenarios:
```
LIST OF TRANSACTIONS FOR EACH SESSION:
---TRANSACTION 651801, ACTIVE 19 sec
LOCK WAIT 2 lock struct(s), heap size 1128, 0 row lock(s)
MariaDB thread id 11486, OS thread handle 139701935003200, query id 2623845 127.0.0.1 root closing tables
DROP TABLE IF EXISTS `sakila`.`staff`
------- TRX HAS BEEN WAITING 19023257 ns FOR THIS LOCK TO BE GRANTED:
TABLE LOCK table `sakila`.`rental` trx id 651801 lock mode X waiting
------------------
---TRANSACTION (0x7f0ef39c2780), not started
0 lock struct(s), heap size 1128, 0 row lock(s)
---TRANSACTION 651786, ACTIVE 19 sec getting list of referencing foreign keys
1 lock struct(s), heap size 1128, 0 row lock(s), undo log entries 11817
MariaDB thread id 11491, OS thread handle 139702133708352, query id 2623861 127.0.0.1 root Opening tables
INSERT INTO `rental` VALUES(11822,"2005-08-17 12:32:39",1317,263,"2005-08-18 12:30:39",2,"2006-02-15 21:30:53")
,(11823,"2005-08-17 12:36:37",2626,352,"2005-08-22 11:10:37",2,"2006-02-15 21:30:53")
,(11824,"2005-08-17 12:37:54",2639,1,"2005-08-19 10:11:54",2,"2006-02-15 21:30:53")
,(11825,"2005-08-17 12:43:30",2656,296,"2005-08-20 15:25:30",1,"2006-02-15 21:30:53")
,(11826,"2005-08-17 12:43:46",1837,536,"2005-08-19 16:59:46",2,"2006-02-15 21:30:53")
,(11827,"2005-08-17 12:44:27",3064,523,"2005-08-24 13:31:27",1,"2006-02-15 21:30:53")
,(11828,"2005-08-17 12:48:28",2593,268,"2005-08-24 09:24:28"
```

As we are over writing, there might be scenarios where referenced tables are being drop when tables are being inserted. 
This should not happen if we test from one database to another empty database or we must empty the database before import. 
We could avoid this scenarios if we could drop the tables at start up, which means that we must know the list of tables to import. This is possible once we implement https://github.com/mydumper/mydumper/issues/1174 when STREAM is used or with the METADATA file when reading from DIRECTORY.

The good thing here is that we detect a locking or sync issue, not an issue in the code, taking into account that we are performing 820 tests now. 

So, the goal is to split the DROPs and the CREATE statements.

davidducos:
After reviewing mydumper code I realized that there is sync mechanism cross all the working thread that force all CREATE TABLE statement to be flushed before starting with the INSERTs. 
This scenario on myloader is rare, as the INSERT on the child table must be right after the parent's CREATE TABLE.  
A way to sort it out is to don't execute INSERTs if any thread is creating a table. Reviewing that now.

davidducos:
Now that we implemented https://github.com/mydumper/mydumper/pull/1201, this should not happening. Merging...

PR #1191, opened 2023-06-30
Multithreaded constraint creation

davidducos:
The performance improvement has been confirmed https://github.com/mydumper/mydumper/discussions/1182


davidducos:
https://github.com/mydumper/mydumper/pull/1206 implements this in a better way

PR #1266, opened 2023-08-15
Adding multicolumn primary key support and refactoring

davidducos:


davidducos:
multicolumn primary key support has been added when we have integer columns. CHAR columns are not supported yet and I had disabled for this branch as it was causing problems after the refactoring

PR #1272, opened 2023-08-22
Using semaphores to prevent excessive local file accumulation during output throttling.

shihuafan:
Make sure the stream output thread begins reading the file before the worker thread advances to the next step. This prevents excessive local file accumulation, which could occupy excessive disk space in scenarios where the output byte stream is blocked or consumed at a slow rate.


davidducos:
Hi @shihuafan,
Why didn't you use a GMutex structure instead of GAsyncQueue?

davidducos (approved):

PR #1278, opened 2023-08-25
fix function: is_table_in_list 

shihuafan:
The input table_names for is_table_in_list do not include the dbname, thus making it unable to match any table.

For example:
`myloader -T` does not work.
`mydumper -T db1.table1 -B db1` will dump all tables in db1.
Warning: In the current version, `mydumper -T db1.table1 -B db1` only dump db1.table1. This is different from the previous version which would dump all tables.

I'm not very familiar with all code, maybe you need to do more verification.

### DIFF
--- a/src/common.c
+++ b/src/common.c
@@ -481,11 +481,16 @@ gboolean m_remove(gchar * directory, const gchar * filename){
   return TRUE;
 }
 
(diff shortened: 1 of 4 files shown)


davidducos:
The errors are related to another issue. When https://github.com/mydumper/mydumper/pull/1266 is merge, this will work

PR #1334, opened 2023-10-31
MyDumper hangs with --stream option if DB is big

midenok:
start_dump() thread first does free_db_table(dbt) then
wait_stream_to_finish(). This is wrong order as
metadata_partial_writer() thread (waited by stream thread) accesses
dbt.

The fix reoders wait_stream_to_finish() so that is done before
free_db_table().

Bug description:

https://jira.mariadb.org/browse/CONTRIB-10
@midenok

### DIFF
--- a/src/mydumper_start_dump.c
+++ b/src/mydumper_start_dump.c
@@ -1205,8 +1205,23 @@ void start_dump() {
   mysql_close(conn);
   g_message("Main connection closed");  
 
-
   wait_close_files();
+
+  g_rename(metadata_partial_filename, metadata_filename);
+  if (stream) {
+    stream_queue_push(NULL, g_strdup(metadata_filename));
+
+    if (exec_command!=NULL){
+      wait_exec_command_to_finish();
+    }else{
+      stream_queue_push(NULL, g_strdup(""));
+      wait_stream_to_finish();
+    }
+    if (no_delete == FALSE && output_directory_param == NULL)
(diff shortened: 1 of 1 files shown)


src/mydumper_start_dump.c
davidducos, line 1: mdfile needs to be close before streaming metadata_filename

midenok: > mdfile needs to be close before streaming metadata_filename

Why?

midenok: @davidducos I can do investigation myself, but if you have some quck knowledge that saves me time, please share!

src/mydumper_start_dump.c
```diff
@@ -1248,26 +1263,11 @@ void start_dump() {
   fclose(mdfile);
```
davidducos, line 1263: Here is where mdfile is being close, but you are streaming metadata before closing. 
The stream backup works like this:
- you open the file and save all the info
- you close the file to guarantee that everything is in there
- you send the file to the stream thread, which is going to reopen, send the file through the stdout and close it again
You should move the close before enqueueing into the stream queue if nothing is been writing in mdfile

midenok: > you close the file to guarantee that everything is in there

Then I guess flushing must be enough. There are chicken and egg problem for doing this after close.

midenok:
Duplicates #1355

PR #1339, opened 2023-11-02
myloader wrong behavior without config file

midenok:
Without config file session variables are not set, but some of them are required for correct restoration (NO_AUTO_VALUE_ON_ZERO and FOREIGN_KEY_CHECKS). The patch sets some reasonable values for myloader session variables, so it could work without configuration file.

Bug description:

https://jira.mariadb.org/browse/CONTRIB-19

The reason for this patch:

Configuration file should not be mandatory for correct work of software. This is user unfriendly! Otherwise you should fail it to run if it doesn't find configuration file. For constant machine-independent data there is /usr/share directory according to FHS, it is wrong to place it into /etc (though it is very correct to have any vars in /etc **above** the default values).

### DIFF
--- a/src/mydumper_common.c
+++ b/src/mydumper_common.c
@@ -548,7 +548,6 @@ void initialize_sql_statement(GString *statement){
   if (is_mysql_like())  {
     if (set_names_statement)
       g_string_printf(statement,"%s;\n",set_names_statement);
-    g_string_append(statement, "/*!40014 SET FOREIGN_KEY_CHECKS=0*/;\n");
     if (!skip_tz) {
       g_string_append(statement, "/*!40103 SET TIME_ZONE='+00:00' */;\n");
     }
@@ -556,8 +555,6 @@ void initialize_sql_statement(GString *statement){
     if (!skip_tz) {
       g_string_printf(statement, "/*!40103 SET TIME_ZONE='+00:00' */;\n");
     }
-  } else {
-    g_string_printf(statement, "SET FOREIGN_KEY_CHECKS=0;\n");
   }
 }
 
--- a/src/myloader.c
+++ b/src/myloader.c
@@ -112,6 +112,21 @@ GHashTable * myloader_initialize_hash_of_session_variables(){
   if (commit_count > 1)
     g_hash_table_insert(set_session_hash,g_strdup("AUTOCOMMIT"),g_strdup("0"));
 
+  const int m= get_major();
+  const int s= get_secondary();
+  const int r= get_revision();
+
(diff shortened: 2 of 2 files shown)


davidducos:
Hi @midenok,
I had been thinking about it too. Why don't we move this to the metadata file. We could add a section like we have in the config file, to avoid setting this on the code or the configuration file. What do you think?

midenok:
> Hi @midenok, I had been thinking about it too. Why don't we move this to the metadata file. We could add a section like we have in the config file, to avoid setting this on the code or the configuration file. What do you think?

Hi! Sorry, I don't understand how to set that in metadata? From the code? So it will be anyway in code. Or am I missing something?

There is nothing wrong with configuration layer. User has ability to override defaults -- great! But down the layer there are defaults user has no access. What is wrong with the code? I would care if there were 20 variables, now I don't see a point to do something else. There are more important problems...

midenok:
> Hi @midenok, I had been thinking about it too. Why don't we move this to the metadata file. We could add a section like we have in the config file, to avoid setting this on the code or the configuration file. What do you think?

Why it is good to be in sql file: you can easily pipe it through mysql client manually.

davidducos:
Hi @midenok 
If a user wants to use a sql file which doesn't have something that he wants, they can execute it in this way:
```
# cat david.sql
select now(6);
# ( echo "select now(6);" ; cat david.sql ) | mysql
now(6)
2023-11-07 18:17:37.121142
now(6)
2023-11-07 18:17:37.121437
```
If a user wants to remove something from a sql file, they must edit every single file.

And on the other hand, if you want to use mysql client is because you take a backup with mysqldump, if you took a backup with mydumper, you should use myloader.

midenok:
@davidducos Sorry, I don't understand the example with concatenation. You can concatenate or cut out text from stream of course (but this is already *not easy*), that's not the point. The point is: some settings are the part of the dump, the dump doesn't work with another settings!

Storing that into metadata will save you some bytes, but this is not as universal as storing into .sql (by the above reason).

davidducos:
Hi @midenok,
I don't want to give you the wrong impression with all my questionings. I have just realized that what we are discussing, has been already fixed on mysqldump with the option `--compact`. As it is implemented right now, --compact is the default and what you are proposing is to implement the non --compact option and set it as default. 

Does it make sense?

midenok:
Closing in favour of #1342

midenok:
Related #1342

davidducos:
@midenok what do you want to do with this PR? can we close it? or are you going to fix?

Issue #1338, opened 2023-11-01
[BUG] myloader fails to restore table that references sequence


midenok:
### Reproduce
```sql
DROP DATABASE IF EXISTS test1;
CREATE DATABASE test1;
USE test1;
CREATE SEQUENCE seq_t_i INCREMENT 5 START WITH 1;
CREATE TABLE t (
  i integer DEFAULT nextval(seq_t_i),
  j integer
);
INSERT INTO t VALUES (1,1),(2,2),(3,3),(4,4);
select * from t order by i;

--let $databases= test1
--source include/mydumper_dump.inc
--source include/myloader_restore_rr.inc
show create sequence test1.seq_t_i;
select * from t order by i;

DROP DATABASE IF EXISTS test1;
```
### Result
```
** (myloader:2556441): WARNING **: 13:03:33.235: Thread 5: Error restoring 1: Table 'test1.seq_t_i' doesn't exist
```
### Cause
It executes table creation while sequence is not yet created.
```sql
/*!40101 SET NAMES binary*/;
/*!40103 SET TIME_ZONE='+00:00' */;
CREATE TABLE `t` (
`i` int(11) DEFAULT nextval(`test1`.`seq_t_i`),
`j` int(11) DEFAULT NULL
) ENGINE=MyISAM DEFAULT CHARSET=latin1 COLLATE=latin1_swedish_ci;

```

midenok:
### Cause
Sequence is created in view_queue which is executed in worker_post_thread() (after tables created and data loaded).

### Fix
database_queue should be refactored into 3 queues: database_queue, sequence_queue, table_schema_queue. sequence_queue must executed before table_schema_queue.

### Code insight

    TODO: this is really strange kind of synchronization: if database is not
    yet created we queue table creation into database queue, then when database
    is created we requeue table creation from database queue into table_queue
    (see set_db_schema_state_to_created()). What for? What kind of performance
    goals this pursuing? All DDL can be done in one thread sequentially without
    all this strange parallelism and that will not be sensible performance
    drawback.

    Another option which fits better current "million queues" design would be to
    execute conf-> queues by sequential stages (parallel in each stage):

      1. database_queue;
      2. table_queue;
      3. data_queue;
      4. post_table_queue;
      etc...

    Now we have several thread pools for that: 1. and 2. is done by
    worker_schema_thread(), 3. is done by loader_thread(), 4. is done by
    worker_post_thread(), etc. This zoopark of thread pools does not add to
[...]

PR #1341, opened 2023-11-06
Fix: myloader fails to restore table that references sequence + stability fixes

midenok:
CREATE TABLE fails if table referencing sequence is restored before sequence. This can happen because sequence restore happens at the very late stage of view restore. To fix this we move sequences restore in between database and table schema restore.

We enqueue sequences to buffer queue database::sequence_queue like it is done for tables with database::queue. And we requeue sequence_queue before tables queue in set_db_schema_state_to_created().  We trigger reque like for tables: after database schema was created. When all sequences processed (sequences_processed == sequences) we requeue tables same way in set_db_schema_state_to_created().

There are 2 stages in INTERMEDIATE_ENDED. When first stage entered we wait until all sequences processed and then initiate for all dbs set_db_schema_state_to_created() which requeues their tables into creating queue.

Bug description:

https://jira.mariadb.org/browse/CONTRIB-25

Fixes mydumper/mydumper#1338


davidducos:
Hi @midenok please, send me a request when a PR is ready to review.

Cheers

midenok:
@davidducos What request? PR is the request itself. It's ready for review.

davidducos:
Hi @midenok, 
As we have all those checks, it is possible that a PR fails and needs some rework. 
That is why, sometimes a PR is not ready to review. 
You can request review from here:
image

Let me know if that works for you and if you are willing to add you as Collaborator to the project.

davidducos:
Hi @midenok 
I think that you are missing a scenario. If you have a sequence loaded after the table, then myloader will not be able to know that there is a sequence that needs to be processed. 
I think that you need to flag the tables that use sequence and hold only those tables, not the entire backup.

midenok:
> Hi @midenok I think that you are missing a scenario. If you have a sequence loaded after the table, then myloader will not be able to know that there is a sequence that needs to be processed.

It counts sequences in `process_metadata_global()` (before reading filesystem).

> I think that you need to flag the tables that use sequence and hold only those tables, not the entire backup.

But this is not possible without proper SQL parser, is it?

davidducos:
> > Hi @midenok I think that you are missing a scenario. If you have a sequence loaded after the table, then myloader will not be able to know that there is a sequence that needs to be processed.
> 
> It counts sequences in `process_metadata_global()` (before reading filesystem).

it is not guarantee when you are using stream.

> 
> > I think that you need to flag the tables that use sequence and hold only those tables, not the entire backup.
> 
> But this is not possible without proper SQL parser, is it?

I think that this query should work:
```
select TABLE_NAME from information_schema.COLUMNS where COLUMN_DEFAULT like 'nextval%' and DATA_TYPE like '%int';
```

midenok:
> > > Hi @midenok I think that you are missing a scenario. If you have a sequence loaded after the table, then myloader will not be able to know that there is a sequence that needs to be processed.
> > 
> > 
> > It counts sequences in `process_metadata_global()` (before reading filesystem).
> 
> it is not guarantee when you are using stream.

What does this mean? It doesn't send metadata first in stream?

> 
> > > I think that you need to flag the tables that use sequence and hold only those tables, not the entire backup.
> > 
> > 
> > But this is not possible without proper SQL parser, is it?
> 
> I think that this query should work:
> 
> ```
> select TABLE_NAME from information_schema.COLUMNS where COLUMN_DEFAULT like 'nextval%' and DATA_TYPE like '%int';
> ```

There is not only nextval(), there is lastval(). There may be more functions in future. There is not only default expressions, but check and virtual column expressions as well. It may be not simple `nextval(x)` expression but any math expression: `(5 + nextval(x) + lastval(y)) * nextval(z)`, etc.

davidducos:
@midenok how does MariaDB version of mysqldump to sort this issue out???

midenok:
> @midenok how does MariaDB version of mysqldump to sort this issue out???

It dumps sequences first, I guess.

davidducos:
Ok, then we need to do on mydumper:
1- Split the table selection, we have to select the sequence with:
```
select TABLE_SCHEMA,TABLE_NAME from information_schema.TABLES where TABLE_TYPE = 'SEQUENCE';
```
And the remaining tables
2- We need to count the sequence. 
3- We are going to send the sequence first because stream needs it asap. Once all the sequence jobs are done, we start sending the tables.
4- We save the amount of sequence on metadata.

On myloader:
- if it is from directory, we need to read the amount of sequence and do not create the tables until we create the tables as you already did on this PR
- if it is stream there shouldn't be any issue, as the sequence are coming first but we could also check on metadata partial the amount of sequence that we need to execute to allow to create tables

@midenok you will need to add support for stream.

midenok:
> Ok, then we need to do on mydumper: 1- Split the table selection, we have to select the sequence with:
> 
> ```
> select TABLE_SCHEMA,TABLE_NAME from information_schema.TABLES where TABLE_TYPE = 'SEQUENCE';
> ```
> 
> And the remaining tables 2- We need to count the sequence. 3- We are going to send the sequence first because stream needs it asap. Once all the sequence jobs are done, we start sending the tables. 4- We save the amount of sequence on metadata.
> 
> On myloader:
> 
> * if it is from directory, we need to read the amount of sequence and do not create the tables until we create the tables as you already did on this PR
> * if it is stream there shouldn't be any issue, as the sequence are coming first but we could also check on metadata partial the amount of sequence that we need to execute to allow to create tables
> 
> @midenok you will need to add support for stream.

Let's do one step at a time! I'm not sure my customer needs stream usage, OTOH everybody will benefit from sequences working in file dump. You should open bug report for the part of streaming. Meanwhile I've been busy with a lot of stability fixes. Here in this patch I added them.

midenok:
Duplicates #1355

PR #1355, opened 2023-11-21
Sequence, ANSI, ORACLE and stability fixes

midenok:
See fix descriptions in commit messages. Sequence fix contans part for dir mode. Stream mode should be fixed as a separate task.

Duplicated by #1341, #1334

Note: there is some setup problem with 2 bionic buildbots.

[Example of myloader trace](https://gist.github.com/midenok/c6ca669ec0f2cbc271c1fd4e80a8736b) Thread descriptor is added only for --debug option


.circleci/config.yml
```diff
@@ -974,18 +974,18 @@ commands:
   prepare_ubuntu_percona57:
     steps:
     - run: sudo percona-release setup -y ps57
-    - run: sudo apt-get install -y libperconaserverclient20 percona-server-client-5.7 libperconaserverclient20-dev
+    - run: sudo apt-get install -y gdb libperconaserverclient20 percona-server-client-5.7 libperconaserverclient20-dev
```
davidducos, line 977: You should not edit the .circleci/config.yml file as it is generated by .circleci/circleci_config_builder.sh

src/myloader.c
```diff
@@ -399,7 +404,7 @@ int main(int argc, char *argv[]) {
 
   struct thread_data t;
   t.thread_id = 0;
-  t.conf = &conf;
+  t.conf = &conf; // TODO: if conf is singleton it must be accessed as global variable
```
davidducos, line 407: On myloader yes! I'm agree with you on this. However, on mydumper it is not.

.circleci/config.yml
midenok: What should be done instead?

src/myloader.c
midenok: This is myloader code. Ok for mydumper.

.circleci/config.yml
midenok: I changed `circleci_config_builder.sh`

davidducos (approved):

midenok:
Thanks!

PR #1360, opened 2023-11-29
Double quote in schema dump fixes

midenok:
For ANSI/ANSI_QUOTES/ORACLE we use double quote as identifier quote character. Dumping views/routines/sequences didn't handle that properly.

SHOW CREATE can work with backticks in any mode. But this is rather a kludge than well-defined behaviour, so this should be avoided.

Bug description:

https://jira.mariadb.org/browse/CONTRIB-34


midenok:
@davidducos Why you creating empty merge commits? Please, I can push my PRs myself.

PR #1381, opened 2024-01-03
Integer sharding deferrer to improve parallelism & some mydumper options

midenok:
--clear, --dirty control how mydumper treats output directory. It doesn't touch dirty out dir by default.
--skip-defer turns off integer sharding deferrer


midenok:
There is a bug in this PR, don't push.

davidducos (approved):

PR #1395, opened 2024-01-26
Myloader hang fix, mydumper lost row fix, cleanups, tracing

midenok:
Lost row fix requires hotfix release!

I think v0.15.2-7 was a bit premature. Lost row happens due to unmasked bug by 640ece1, so now  v0.15.2-7 is spoiled.

There should be some stabilization period (a month) after so many changes before issuing a release.

Now as CircleCI blocked me it doesn't even run tests for my PRs. Tests passed locally.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,4 +15,4 @@ docs/man/
 export-*
 mydumper
 myloader
-
+sakila-db*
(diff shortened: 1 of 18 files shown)


davidducos (approved):

midenok:
@davidducos There is one more myloader hang known. The fix is undergoing.

PR #1396, opened 2024-01-26
--overwrite-tables hang fix 3

midenok:


midenok:
This is very stable point. test_mydumper.sh is running repeatedly since yesterday without hang.

davidducos (approved):

Issue #1292, opened 2023-09-06
[FEATURE REQUEST] --skip-checksum on myloader


utdrmac:
**Describe the Feature Request**
There is no option to skip the checksum process when running myloader. Checksum of tables happens even when using `--no-data` and there should be an option to skip the checksums.

PR #1397, opened 2024-01-29
Checking improvements

midenok:
New options:

myloader --checksum [fail|warn|skip]: control how to treat checksums in dump (made with --checksum-all)

- Default is "fail" which means stop myloader on first checksum
mismatch;

- "warn" works like before: print warning;

- "skip" makes myloader as quick as without checksums, i.e. it skips
checksums calculation;

mydumper --check-row-count: Rows count check when making dump

When enabled, row count is done by SELECT COUNT(*)
which is slower than EXPLAIN SELECT but accurate. If dumped row count is different mydumper fails with error message.

### DIFF
--- a/src/common.c
+++ b/src/common.c
@@ -31,6 +31,7 @@
 
 GAsyncQueue *stream_queue = NULL;
 gboolean skip_defer= FALSE;
+gboolean check_row_count= FALSE;
 
 /*
 const char *usr_bin_zstd_cmd[] = {"/usr/bin/zstd", "-c", NULL};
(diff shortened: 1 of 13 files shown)


davidducos:
@midenok, why are you removing rows_in_sts (rows in show table status)? it is useful to perform calculations when PK is integer.

midenok:
@davidducos It was not used, it is dead code. AFAICS you replaced them by EXPLAIN SELECT.

davidducos (approved):

davidducos:
#1292

PR #1414, opened 2024-02-22
Disabling split by varchar

davidducos:
I decided that we are going to disable the chunks by varchars.
It was not implemented correctly. 
It was a good exercise, but it is not mature enough. 

We need to develop a better understanding of the use cases where it will be useful. For instance, when binary is used or when UTF8MB4, and what is the best approach to split the chunks.

### DIFF
--- a/src/mydumper_chunks.c
+++ b/src/mydumper_chunks.c
@@ -217,11 +217,9 @@ struct chunk_step_item * initialize_chunk_step_item (MYSQL *conn, struct db_tabl
       case MYSQL_TYPE_STRING:
       case MYSQL_TYPE_VAR_STRING:
 
-/*
         if (minmax)
           mysql_free_result(minmax);
         return new_none_chunk_step();
-*/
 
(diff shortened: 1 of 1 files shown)


midenok:
Any evidences it doesn't work correctly?

davidducos:
@midenok, 
when you mix digits and upper and lower characters, you will might end up exporting same chunks on different moments. This was not detected before due to checksum bugs.

midenok:
What is the consequence of this? Duplicate PK?

davidducos:
Hi @midenok, yes, data was being exported twice (or more) and in some scenarios, it was not being exported due to invalid characters. 
I checked mysqlsh strategy and I don't like it as it is doing a SELECT MIN,MAX LIMIT N, which might take a similar amount of time of executing the SELECT to extract the data. I think that we can do better, but we need to take into consideration the character set and collation to do not leave gaps and make it performant.
The problem to solve is how we determine the min and max when we create a new chunk, taking into considerations the estimation of rows to export. 
Taking into account that we have utf8mb4 now, we need to plan a good strategy... my first implementation was good for understanding this and I learn a lot, but it was not going to scale and it didn't work on all use cases.

PR #1455, opened 2024-03-27
Fix the issue of SQL execution error when the database name contains characters that need to be escaped.

jackytsu:


jackytsu:
> This show me that we have a bigger issue, as in another portion of code we have:
> 
> ```
>     char *field_name= backtick_protect(row[0]);
>     gchar *tb = g_strdup_printf("`%s`", field_name);
> ```
> 
> In the line that you fixed and on this one we need to use identifier_quote_character_str and we need to create protect function for identifier_quote_character_str.

Yeah, that's another problem, but I haven't  meet it, yet. ;)

PR #1456, opened 2024-03-27
some optimisation when using --tables-list

shihuafan:
skip dump mysql.show_log .... when using `mydumper --tables-list='mysql.%'`


davidducos:
Hi @shihuafan , please, request a review when you are done.

src/mydumper_start_dump.c
```diff
@@ -822,62 +836,44 @@ void send_lock_all_tables(MYSQL *conn){
...
-            if (g_ascii_strcasecmp(tables[i], row[1]) == 0)
-              table_found = 1;
-          if (!table_found)
-              lock = 0;
-        }
-        if (lock && tables_skiplist_file && check_skiplist(row[0], row[1]))
+        // no need to check if the tb exists in the tables.
+        if (tables_skiplist_file && check_skiplist(dt[0], row[0]))
           continue;
-        if (lock && !eval_regex(row[0], row[1]))
```
davidducos, line 863: @shihuafan please, review my last commit about this change

shihuafan: ok

shihuafan:
@davidducos I'm done, please review
here are some modifications about tables-list parameter
1. skip special tables when search tables, such as mysql.slow_log
2. when lock-all-tables, use sql to determine which tables need locking
3. when tables-list parameter exists, do not execute the
create_job_to_dump_schema。Older versions will execute this function, but will discard all tables that are not in the tables-list.
4. in_table_list adds schema matching for mysql like, (but this function is useless when dumping, it only works in myloader)

PR #1468, opened 2024-04-09
[bug] Fix --max-threads-per-table not working with myloader

delissonjunio:
In the latest mydumper (compiled from source), `--max-threads-per-table` is not working with `myloader`. It just creates `max-threads-per-table` threads for the first table and sits there, not using the other `--threads` it should.

It looks like that's caused by a missed iteration step when looking for tables that can be worked on.


inderjit-cloud (approved):
Tested this fix and I can see the threads actively running per table.

davidducos:
Hi @delissonjunio,
Thanks for notice this bug, I also found when I was working on https://github.com/mydumper/mydumper/pull/1474. 
You will find there the same code that you added and a lot of changes. Take into account that --max-threads-per-table is still take into account but not in the control_job. 
This is why, I will be closing this. Thank you again for reviewing and found it!

PR #1494, opened 2024-05-13
Adding --success-on-1266 to avoid CRITICAL on engine substitution

davidducos:


davidducos:
While I was implementing this I realized about a problem. mysql_errno and mysql_error only show the first error on the list. It is not possible for us to detect if it is ok to continue when engine substitution has happened. I also found that multiple errors lead to this scenario. We need to implement an option for a list of errors and not just 1 case. The engine substitution will depend if NO_ENGINE_SUBSTITUTION has been used on SQL_MODE.

PR #1510, opened 2024-05-22
Adding --metadata-refresh-interval option

davidducos:


davidducos:
In you create a backup and then edit the metadata to get something like this:
```
# Started dump at: 2024-05-22 12:40:25
[config]
quote_character = BACKTICK

[myloader_session_variables]
SQL_MODE='NO_AUTO_VALUE_ON_ZERO,ONLY_FULL_GROUP_BY,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,NO_ENGINE_SUBSTITUTION' /*!40101

[master]
# Channel_Name = '' # It can be use to setup replication FOR CHANNEL
File = binlog.000026
Position = 30988347
Executed_Gtid_Set =
```
And then you execute:
```
for i in $(seq 1 100000); do echo "[\`sbtest\`.\`sbtest${i}\`]
real_table_name=sbtest${i}
rows = 10" ; done >> metadata
```
You will get a metadata file which will contain 100k tables to load in memory. 

Then if you test with different values you will get:
```
root@ubuntu-jammy:~/issue_1487/mydumper# time ./myloader -d ~/issue_last_test/mydumper/data -o --verbose=3 -B sbtest2 -L log --metadata-refresh-interval 10000

real	0m1.642s
user	0m1.570s
sys	0m0.070s

root@ubuntu-jammy:~/issue_1487/mydumper# time ./myloader -d ~/issue_last_test/mydumper/data -o --verbose=3 -B sbtest2 -L log --metadata-refresh-interval 1000

real	0m2.211s
user	0m2.118s
sys	0m0.092s

root@ubuntu-jammy:~/issue_1487/mydumper# time ./myloader -d ~/issue_last_test/mydumper/data -o --verbose=3 -B sbtest2 -L log --metadata-refresh-interval 100

real	0m7.574s
user	0m7.411s
sys	0m0.140s

root@ubuntu-jammy:~/issue_1487/mydumper# time ./myloader -d ~/issue_last_test/mydumper/data -o --verbose=3 -B sbtest2 -L log --metadata-refresh-interval 10
d^\Quit (core dumped)

real	0m43.509s
user	0m42.928s
sys	0m0.364s
```
I cancelled the last one as I didn't want to wait, as I think that you get the point. Default is 100. 

The big question here is: Do we need to sort the tables based on rows? Honestly, we need to test it with the latest release as it is using a smarter way to import the data as it is using multiple connection and a connection pooling. I don't want to end up in an scenario where it will depend on the amount or size of your tables. The whole idea of sorting the tables is when --innodb-optimize-keys and you have large tables which should be done at first.

PR #1574, opened 2024-07-24
Implemeting FORMAT MySQLDump for clickhouse

davidducos:


davidducos:
We are not going to continue with this. However, we are going to create a ticket to be able to extract data from Clickhouse as currently this is not possible.

PR #1605, opened 2024-09-17
Adding ENCRYPTION='N' when temporary table on memory is created for views

davidducos:


davidducos:
Caused a bug fixed on https://github.com/mydumper/mydumper/pull/1623/commits/ab6b65ae42436a3ed4bdcf9606ed9026cb7f20bc

PR #1651, opened 2024-10-23
Write buffer added and table structure used has changed

davidducos:


davidducos:
This was implemented on other branches and I found no need to implement the main use case for this PR.

PR #1729, opened 2025-01-14
add note about versioning to readme

michaelglass:
this is as much a question about versioning as a minor documentation addendum. Please please close without explanation if you think the addendum isn't helpful.

Is it true that even patches are pre-releases?


davidducos (approved):

davidducos:
Hi @michaelglass, the major version is still 0, I hope to release v1 this year and we will be able to change this.

PR #1775, opened 2025-02-18
Moving to PCRE2 and adding regex to masquerade 

davidducos:
Moving PCRE to PCRE2
Adding regex to masquerade

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -140,11 +140,11 @@ MESSAGE(STATUS "------------------------------------------------")
 MESSAGE(STATUS)
 
 
-get_cmake_property(_variableNames VARIABLES)
-list (SORT _variableNames)
-foreach (_variableName ${_variableNames})
-    message(STATUS "${_variableName}=${${_variableName}}")
-endforeach()
+#get_cmake_property(_variableNames VARIABLES)
+#list (SORT _variableNames)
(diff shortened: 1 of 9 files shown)


davidducos:
We made multiple changes, now the basic functions receives the same parameters:
- WITH_MEM: which does the same that _with_mem functions
- REPLACE_NULL: in case that a null value is found, it will build a new value for that row.
- UNIQUE: it will record the values added and do not allow that the same value be assigned to different rows. 
- MAX_LENGTH <NUMBER>: it defines the max length of the row.

We also added tag <regex> which receives an string surrounded by simple quotes which will be the regular explession taht will be replace bu the next tag or constant.
This are simple examples that works:

```
[`sakila`.`customer`]
`store_id`=random_int WITH_MEM REPLACE_NULL UNIQUE MAX_LENGTH 2
`email`=random_format <regex '.{3}$'>'com'
```

In that case the `@sakilacustomer.org` will be change to `@sakilacustomer.com` but you could have a different output like:
```
`email`=random_format <regex '.{3}$'><string 3>
```
Which output could be: `@sakilacustomer.djk`,  `@sakilacustomer.pwn`, or any other random string on the last 3 characters.

davidducos:
<file> is not supported yet

PR #1825, opened 2025-04-07
Refactoring multicolumn

davidducos:
Previous this PR, when multicolumn was possible, the chunk size on the first level wat set statically to 1.
Currently, the chunk size is dynamic. The second level is created when the calculations allow it based on the rows from the the explain.
This change should be affecting to multicolumn tables only. Some tuning or review might be needed in the future.


davidducos:
This should be fixing bug reported in #1808

PR #1848, opened 2025-04-23
Fixing multicolumn bug

davidducos:


davidducos:
This is related to #1808

PR #1865, opened 2025-05-08
Fix on full table scan when second column was string on multicolumn PK

davidducos:


davidducos:
Related to #1808

davidducos:
Test added https://github.com/mydumper/mydumper/pull/1869

PR #1958, opened 2025-07-24
Adding --server-version to avoid automatic server detection

davidducos:
MyDumper detected the product and version of the servers that it was connecting automatically. However, this might be causing issue when some specifics products were being used, like AWS, and others. 
With --server-version you will be able to set the product and version like this: "<PRODUCT> - <VERSION>" , where product will be the know products: MySQL, Percona, MariaDB, Source, Dolt and TiDB and the VERSION will be 3 numbers delimited by dots, like: 5.7.30, 8.0.24, 10.6.34, etc

### DIFF
--- a/src/common_options.c
+++ b/src/common_options.c
@@ -65,6 +65,8 @@ gint source_data=0;
 gchar *throttle_variable=NULL;
 guint throttle_value=0;
 
+gchar *server_version_arg=NULL;
+
 GOptionEntry common_entries[] = {
     {"threads", 't', 0, G_OPTION_ARG_INT, &num_threads,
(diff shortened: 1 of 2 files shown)


DaviMedrade:
I'm commenting this here instead of in that bug I opened earlier because it seems to me like a small, self-contained issue that is mostly unrelated to that. I don't even know if GitHub lets you edit the message of a PR that has been merged for this long, but still…

Right where you show a template of the expected syntax, the words are missing. From what comes after, I'm guessing it should be:

… you will be able to set the product and version like this: "product-VERSION", where product…

But, at least to me (Linux desktop, I checked with both Firefox and Chrome) it reads:

… you will be able to set the product and version like this: " - " , where product …

PR #2004, opened 2025-09-01
Setting correct rand size and ignore case when masquerade based on file

davidducos:


davidducos:
Ignoring checks as this is not being tested.

PR #2031, opened 2025-09-17
Adding SAFE_NO_LOCK sync thread mode

davidducos:
Thanks Fer Mattera! This was a really good idea!

Running this in console 1:
```
sysbench /usr/share/sysbench/oltp_read_write.lua --table-size=10000 --tables=1 --mysql-user=root run --create_secondary=off --mysql-host=${MYSQL_HOST} --time=100 --report-interval=1
```

You can get this output during the backup when using NO_LOCK:
```
circleci@f7116bec80a9:~/issue_1694/mydumper$ ./mydumper -u root -B sbtest -v 4 -o data --clear --sync-thread-lock-mode=NO_LOCK -L log ; grep consis log
2025-09-17 11:33:29 [INFO] - Executing in NO_LOCK mode, we are not able to ensure that backup will be consistent
2025-09-17 11:33:29 [WARNING] - There are differences in the binlog position at the beginning of the backup and after syncing threads, so we cannot guarantee the backup to be consistent due to the use of NO_LOCK. Continues anyway, use SAFE_NO_LOCK otherwise.
```
And this when using SAFE_NO_LOCK
```
circleci@f7116bec80a9:~/issue_1694/mydumper$ ./mydumper -u root -B sbtest -v 4 -o data --clear --sync-thread-lock-mode=SAFE_NO_LOCK -L log ; grep consis log
** (mydumper:26257): ERROR **: 11:33:44.160: There are differences in the binlog position at the beginning of the backup and after syncing threads, so we cannot guarantee the backup to be consistent. Stopping backup due to the use of SAFE_NO_LOCK.
```


davidducos:
You should consider using `--sync-thread-lock-mode=SAFE_NO_LOCK` and `--replica-data` if you are taking backup from a replica as --replica-data will stop replication to take the replication position which allows mydumper to sync the threads in a safe fashion. 
This is really useful if you want to take consistent backups from a RDS instance, just remember to use `--source-control-command=aws`

Issue #2075, opened 2025-11-11
Change the way we create temporary files


davidducos:
We should do create our temporary files in this way:
```
MySQL uses a trick that has been part of POSIX systems forever. It opens the temp file, and immediately unlinks it. Therefore it's not viewable in any directory listing. But POSIX systems like UNIX and Linux shouldn't actually remove an unlinked file while a process has an open file handle to it. So once the query using the temp table finishes, it will close the file handle, and then the OS will automatically remove the file and free the storage it was using.

This is generally better than requiring the server code remember to remove the tempfile when it's done with it. It also accounts for like the thread terminating, or mysqld crashing. At least it won't leave stale temp files littering your filesystem.

```
Reference: https://stackoverflow.com/a/66174035

davidducos:
This enhancement was harder to implement than what I initially thought.
Because there are 3 use cases where we need fifo:
1- when we use compression
2- when we use stream
3- when we use load data

In order to make it right, we need to create the fifo and unlink it asap. However, when we use compression or load data, this is not so simple, as the file needs to be open first for the process or they are not going to have access to it. 

And we also have the combinations of them!! as if you only use compression, it is not an issue, if you use only load data, there is no issue... but the issues arise when use use compression and load data, and any case with stream. 

I think that we should be creating the fifo and load data/the decompression method during the stream stage/process if needed, instead of reopening the file. This might cause of process to be created but wait for the loaders to read from those fifo files. 
This might be awkward for some users, as you might see hundreds of process waiting, the good thing here is that it will be managed by the OS, so once more time, we rely on the OS.

davidducos:
Actually, the main problem here is load data, as we need to keep the file until mysql library opens it. So, even if we use a fifo during load data, and a failure occurs, the file will remain there. 

So, using unlink when stream is not a problem as we are sharing the fd, using unlink when compression is not an issue as the uncompress thread will unlink it, but any case that uses load data, we are not going to be able to unlink the load data files.

PR #2119, opened 2025-12-01
Feature: Unlink FIFO immediately after both ends connect (Issue #2075)

adwait1290:
# Feature: Unlink FIFO Immediately After Both Ends Connect (Issue #2075)

## Summary

Implements automatic FIFO cleanup for crash safety by adopting MySQL's approach to temporary file handling: create file, unlink immediately, keep using via file descriptor.

**Related Issues:** #2075

### Current Behavior

When myloader uses FIFOs for decompression or LOAD DATA operations:
1. `mkfifo()` creates a named pipe on the filesystem
2. Processes open the FIFO and transfer data
3. `remove()` is called only after the operation completes

If myloader crashes mid-operation, stale FIFO files remain on the filesystem in `/tmp` or `--fifodir`. Users must manually clean these up.

### User Impact

```
$ ls /tmp/myloader_*/
# Stale FIFOs from crashed restores
db.table-schema.sql.zst
db.table.sql.zst
...
```

### POSIX Unlink-After-Open Pattern

POSIX systems don't actually remove an unlinked file while a process has an open file handle to it. Once both ends of a FIFO are connected via file descriptors, the named pipe can be unlinked from the filesystem while the pipe itself continues working.

Reference: https://stackoverflow.com/a/66174035

#### 1. Decompression FIFOs (`myl_open`)

After `fopen()` succeeds (both myloader and subprocess have the FIFO open), immediately unlink:

```c
// src/myloader/myloader_process.c, after line 92 (file=g_fopen)
if (file != NULL) {
    remove(fifoname);  // Unlink immediately - pipe stays usable via fd
}
```

Remove redundant cleanup in `myl_close()`:

```c
// src/myloader/myloader_process.c, line 137
// Removed: remove(f->stdout_filename);
// FIFO already unlinked in myl_open()
```

#### 2. LOAD DATA FIFOs (`myloader_restore.c`)

LOAD DATA is more complex because MySQL opens the FIFO asynchronously when processing the statement. We use a background thread to monitor and unlink:

```c
// src/myloader/myloader_restore.c
struct fifo_unlink_data {
    gchar *fifo_filename;
    int child_pid;
};

static void *load_data_fifo_unlink_thread(void *data) {
    struct fifo_unlink_data *fud = data;

    // Wait for MySQL to process LOAD DATA and open the FIFO
    g_usleep(100000);  // 100ms

    // Check if subprocess is still running (means MySQL connected)
    int status = 0;
    if (waitpid(fud->child_pid, &status, WNOHANG) == 0) {
        // Data flowing = safe to unlink
        remove(fud->fifo_filename);
    }

    g_free(fud->fifo_filename);
    g_free(fud);
    return NULL;
}
```

The background thread is spawned after queuing the LOAD DATA statement:

```c
// After g_async_queue_push(cd->queue->restore, ir)
if (is_fifo) {
    schedule_load_data_fifo_unlink(load_data_fifo_filename, load_data_child_pid);
}
```

The existing `m_remove0()` cleanup at the end remains as a fallback.

### Test Scenarios

1. **Compressed file restore**
   ```bash
   mydumper -d /dump --compress
   myloader -d /dump --threads 16
   ```

2. **LOAD DATA restore**
   ```bash
   mydumper -d /dump --load-data --compress
   myloader -d /dump --threads 16
   ```

3. **Crash recovery**
   ```bash
   myloader -d /dump --threads 16 &
   sleep 5
   kill -9 $!
   ls /tmp/myloader_*/  # Should be empty
   ```

| Scenario | Before | After |
|----------|--------|-------|
| Normal completion | Works | Works |
| Crash during decompression | Stale FIFOs | Auto-cleaned |
| Crash during LOAD DATA | Stale FIFOs | Auto-cleaned (usually) |

## Files Changed

| File | Changes |
|------|---------|
| `src/myloader/myloader_process.c` | Unlink FIFO after fopen in `myl_open()`, remove redundant cleanup in `myl_close()` |
| `src/myloader/myloader_restore.c` | Add background thread for LOAD DATA FIFO cleanup, track subprocess pid |

### LOAD DATA Timing

The LOAD DATA cleanup uses a 100ms delay to give MySQL time to open the FIFO. If MySQL is slow to process the statement, the unlink might happen before MySQL opens the file, causing LOAD DATA to fail with "file not found". In this case, the existing `m_remove0()` fallback still cleans up after MySQL finishes.

This is a best-effort improvement. For guaranteed cleanup, users can configure a shorter timeout or rely on the existing fallback mechanism.

### Stream Mode

Stream mode sharing file descriptors directly may need additional consideration. This PR focuses on the standard decompression and LOAD DATA use cases.

## Backward Compatibility

- No CLI changes
- No dump format changes
- Existing `m_remove0()` cleanup remains as fallback
- Users see no behavioral difference (only more reliable cleanup)

### DIFF
--- a/src/myloader/myloader_process.c
+++ b/src/myloader/myloader_process.c
@@ -120,6 +120,15 @@ FILE * myl_open(char *filename, const char *type){
 
     child_proc = execute_file_per_thread(filename, fifoname, command);
     file=g_fopen(fifoname,type);
+
+    // Issue #2075: Unlink FIFO immediately after both ends are connected.
+    // Once fopen succeeds, both our process and the subprocess have the FIFO open.
+    // The pipe remains usable through the open file descriptors.
+    // This ensures automatic cleanup on crash (no stale FIFOs left behind).
+    if (file != NULL) {
+      remove(fifoname);
+    }
+
     g_mutex_lock(fifo_table_mutex);
     struct fifo *f=g_hash_table_lookup(fifo_hash,file);
     if (f!=NULL){
@@ -166,7 +175,8 @@ void myl_close(const char *filename, FILE *file, gboolean rm){
     g_mutex_unlock(f->mutex);
     g_mutex_unlock(fifo_table_mutex);
 
-    remove(f->stdout_filename);
+    // Issue #2075: FIFO is already unlinked in myl_open() after both ends connect.
+    // No need to remove here - the FIFO name no longer exists on filesystem.
 
     // Release decompressor slot
     if (f->uses_decompressor){
--- a/src/myloader/myloader_restore.c
+++ b/src/myloader/myloader_restore.c
@@ -22,6 +22,9 @@
 #include <stdio.h>
 #include <string.h>
 #include <errno.h>
+#include <fcntl.h>
+#include <poll.h>
+#include <sys/wait.h>
 
 #include "myloader.h"
 #include "myloader_common.h"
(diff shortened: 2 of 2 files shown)


davidducos:
Hi @adwait1290 , there is another conflict, can you take it?

davidducos (approved):

davidducos:
Hi @adwait1290 ,

I found that in some case the value in g_usleep might not be enough. So, I will be tacking this issue with a non-blocking command execution. I will be pushing the PR soon.

PR #2193, opened 2026-02-23
Fixing inconrrect detection of Google Cloud for MySQL

davidducos:
The MySQL instance was not being correctly detected on Google Cloud. ~~I think that it is better to consider it as a normal mysql instance, but this might change in the future, as we might need to set this variables: --skip-ddl-locks --table-engine-for-view-dependency=InnoDB~~
We added the Google type of server and it will work as MySQL. However, we fixed the configuration to take into account particular issues that could arise during dumps and/or restores.


vinoths-1:
@davidducos Thank you for fixing the issue. Could you please let me know which release will include this fix?

rimonghobryal:
@davidducos  Thank you for fixing the issue. Could you please let me know which release will include this fix?

PR #2223, opened 2026-03-21
Fix stream-directory restore ordering in myloader

AndrzejW2016:
## Summary
- fix sorted queue comparators to return proper `<0/0/>0` values instead of booleans
- serialize `process_file_type` when restoring a stream-to-directory dump containing `metadata.partial.*`
- hold the directory metadata release until the full partial-metadata chain has been processed
- ensure graceful shutdown can finish writing resume state
- reduce normal-log noise for filtered tables during selective restore

## Why
Restoring a directory produced from `--stream=TRADITIONAL` was unstable in `myloader`.
In local reproduction this showed up as either early `malloc(): unaligned tcache chunk detected` crashes, inconsistent restore ordering, or shutdown/resume hangs.

Two initial issues stood out:
1. Several comparators used with GLib sorted queues returned only `0/1`, which violates the comparator contract and can lead to unstable ordering.
2. Directory restores containing `metadata.partial.*` could start schema/data scheduling before the partial metadata chain had finished rebuilding the table state.

A second issue showed up in shutdown handling:
- during signal-driven shutdown, resume generation could block because the queue consumer was waiting for its sentinel even though graceful shutdown had already started.

There is also an operational usability problem for selective restore (`-T db.table`) against very large flat stream-to-directory dumps. In our case the dump directory contains on the order of tens of thousands of files in a single directory. Without these fixes, targeted restore becomes difficult to use because:
- restore ordering becomes unstable for stream-directory metadata
- shutdown/resume can hang after signal handling
- normal logs get flooded by skipped objects unrelated to the selected table

## Validation
- rebuilt `myloader` locally with `make myloader`
- confirmed sorted file-type comparisons now log the expected ordering, e.g. `DATA <> SCHEMA_TABLE = 1`
- confirmed a schema file that had previously failed on a dirty target now restores correctly on a clean target DB
- confirmed graceful shutdown now reaches `Shutting down gracefully completed.` and writes `resume` instead of stalling after `Writing resume.partial file`

## Notes
This is intentionally scoped to the ordering, filtering, and shutdown issues above. There is still separate restore-path behavior worth investigating further for this dump format.

### DIFF
--- a/src/myloader/myloader_directory.c
+++ b/src/myloader/myloader_directory.c
@@ -29,6 +29,58 @@
 
 GAsyncQueue *metadata_sync_queue=NULL;
 
+static gboolean filename_matches_selected_tables(const gchar *filename){
+  if (tables == NULL || filename == NULL)
+    return FALSE;
+
+  gchar *database = NULL;
+  gchar *table = NULL;
+
+  if (m_filename_has_suffix(filename, "-schema-view.sql")){
+    get_database_table_from_file(filename, "-schema-view", &database, &table);
+  } else if (m_filename_has_suffix(filename, "-schema-sequence.sql")){
+    get_database_table_from_file(filename, "-schema-sequence", &database, &table);
+  } else if (m_filename_has_suffix(filename, "-schema-triggers.sql")){
+    get_database_table_from_file(filename, "-schema-triggers", &database, &table);
+  } else if (m_filename_has_suffix(filename, "-schema-post.sql")){
+    get_database_table_from_file(filename, "-schema-post", &database, &table);
+  } else if (m_filename_has_suffix(filename, "-schema.sql")){
+    get_database_table_from_file(filename, "-schema", &database, &table);
+  } else if (m_filename_has_suffix(filename, ".sql")){
+    gchar **split = g_strsplit(filename, ".", 4);
+    if (g_strv_length(split) >= 2){
+      database = g_strdup(split[0]);
+      table = g_strdup(split[1]);
+    }
+    g_strfreev(split);
+  }
+
+  gboolean match = FALSE;
+  if (database != NULL && table != NULL)
+    match = is_table_in_list(database, table, tables);
+
+  g_free(database);
+  g_free(table);
+  return match;
+}
+
+static gboolean should_queue_filename(const gchar *filename){
+  if (filename == NULL)
+    return FALSE;
+
+  if (!strcmp(filename, "metadata"))
+    return FALSE;
+
+  if (tables == NULL)
+    return TRUE;
+
+  if (!strcmp(filename, "all-schema-create-tablespace.sql")) {
+    return TRUE;
+  }
+
+  return filename_matches_selected_tables(filename);
+}
+
 void initialize_directory(){
   metadata_sync_queue=g_async_queue_new();
 }
@@ -82,11 +134,11 @@ void *process_directory(struct configuration *conf){
   }else{
     GDir *dir = g_dir_open(directory, 0, &error);
     while ((filename = g_dir_read_name(dir))){
-      if (strcmp(filename, "metadata"))
+      if (should_queue_filename(filename))
         process_filename_push(filename);
     }
+    g_dir_close(dir);
   }
   process_filename_queue_end();
   return NULL;
 }
-
--- a/src/myloader/myloader_process.c
+++ b/src/myloader/myloader_process.c
@@ -45,6 +45,7 @@ GMutex *fifo_table_mutex=NULL;
 struct configuration *_conf;
 extern gboolean schema_sequence_fix;
 GAsyncQueue *partial_metadata_queue = NULL;
+static GRecMutex *metadata_process_mutex = NULL;
 
 // Decompression throttle: limit concurrent decompressor processes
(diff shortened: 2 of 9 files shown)


AndrzejW2016:
Additional context from local reproduction:

Selective restore (`-T db.table`) against a very large flat stream-to-directory dump is especially problematic here. In our case the dump directory contains on the order of tens of thousands of files in a single directory, and without these ordering / shutdown / log-noise fixes `myloader` becomes very hard to use for targeted restore work:

- restore ordering becomes unstable for stream-directory metadata
- shutdown/resume can hang after signal handling
- normal logs get flooded by skipped objects unrelated to the selected table

That makes selective restore on large single-directory dumps operationally expensive even when the user explicitly points at one table.

AndrzejW2016:
Added another fix on top of this branch: when `-T/--tables-list` is used against a large flat dump directory, `process_directory()` now queues only files that belong to the selected tables plus required global files.

I verified it locally on a stream-directory dump with a small selective-restore scenario.

Before this, the selective restore kept stalling behind unrelated directory entries and metadata handling. With the latest patchset it completed successfully and restored all requested tables.

One remaining warning in that test is the existing schema-create checksum mismatch for the database-level checksum, but the selective restore itself completed correctly.

frankii91:
This is what I've been waiting for, finally an efficient restore version will work.

AndrzejW2016:
Added another local performance-focused patch on top of this branch:

- parallelized the process_filename stage by using multiple filename workers with a single file_type completion signal
- changed data scheduling to rely on ready_table_queue while files are still being enqueued
- enqueue ready tables immediately when schema becomes CREATED or when data jobs are appended

Why this matters:

- large flat dump directories should not be bottlenecked by a single-threaded filename classification stage
- the loader scheduler should not repeatedly fall back to O(n) full table-list scans while jobs are still arriving
- tables that are already runnable should become dispatchable immediately instead of waiting for another broad scheduler sweep

AndrzejW2016:
One follow-up optimization that may be worth considering separately is a stronger parallel directory-scan / filename-pipeline model for very large flat dump directories.\n\nWhy it could matter:\n- when a dump directory contains tens of thousands of files in a single level, startup latency is no longer negligible on its own\n- this is especially noticeable when many files are tiny, sparse, or operationally unimportant compared to the few files that actually drive restore progress\n- in that shape of dump, a single-threaded or lightly threaded filename stage can delay how quickly the restore reaches runnable work, even before the main data/index tail becomes the bottleneck\n\nWhere it helps most:\n- full restore from very large flat directories\n- dumps with a high file-count / low useful-work ratio\n- environments where filesystem traversal itself is expensive enough to show up in wall-clock time\n\nThis is separate from the end-game issue of a few large remaining tables or long ALTER TABLE work. It would mainly improve the front-loaded cost of turning a large directory into runnable restore jobs.

AndrzejW2016:
Additional data point from the backup shape that triggered this discussion:\n\n- files: 27098\n- empty files: 17193\n- total size: 6.3G\n\nThat suggests there may also be room for improvement on the dump-production side, not only in restore scheduling.\n\nA separate design direction worth considering is reducing or coalescing empty/sparse output files during backup generation, because a very high file-count with many empty files makes restore pay a filesystem and classification cost that does not translate into useful restore work.\n\nIn other words, part of the restore overhead here may be a symptom of the dump layout itself:\n- many empty files still need to be traversed, classified, queued, and filtered\n- this increases startup overhead and can amplify single-directory scaling problems\n\nSo there may be two complementary levers:\n1. make restore-side directory scanning / filename processing more scalable\n2. reduce the amount of empty or near-empty output artifacts produced by the backup format in the first place

davidducos:
@

> Additional data point from the backup shape that triggered this discussion:\n\n- files: 27098\n- empty files: 17193\n- total size: 6.3G\n\nThat suggests there may also be room for improvement on the dump-production side, not only in restore scheduling.\n\nA separate design direction worth considering is reducing or coalescing empty/sparse output files during backup generation, because a very high file-count with many empty files makes restore pay a filesystem and classification cost that does not translate into useful restore work.\n\nIn other words, part of the restore overhead here may be a symptom of the dump layout itself:\n- many empty files still need to be traversed, classified, queued, and filtered\n- this increases startup overhead and can amplify single-directory scaling problems\n\nSo there may be two complementary levers:\n1. make restore-side directory scanning / filename processing more scalable\n2. reduce the amount of empty or near-empty output artifacts produced by the backup format in the first place

You will not have empty files if you do not use --build-empty-files when taking the backup. If you have empty files even without --build-empty-files, then report a bug with a reproducible test case and I will review it. 

I think that it might be possible to have a list of directories to import.

davidducos (approved):

AndrzejW2016:
I took another pass on this with your review comment in mind.

The earlier version kept table-list filtering in more than one place. In this follow-up I changed it so the filename-level filtering goes through shared logic instead of duplicating the check separately during directory scan.

I also kept `.dat` files aligned with the same decision, because they are needed for `LOAD DATA` restores and should not be filtered differently from the matching `.sql` files.

I'm still fairly new to this part of the codebase, so I tried to keep the change small and focused on the review concern rather than broadening it further.

Follow-up PR: #2230

PR #2226, opened 2026-03-21
Release idle myloader connections during tail wait

AndrzejW2016:
## Summary
- release idle connections from the restore connection pool while loader workers are waiting near the tail of the restore
- lazily recreate a pooled `MYSQL` handle when that slot is borrowed again
- keep this isolated from the directory scan / scheduling work in #2223

## Why
During local restore tests, `myloader` kept most DB sessions open until the very end even when only a few loader workers still had work left.
This was visible in `PROCESSLIST` during selective restores of large tables and unnecessarily held DB resources for idle worker slots.

## Implementation
- add `release_idle_connection_if_possible()` to close one currently free pooled connection
- call it from the loader control path when all jobs are already enqueued, there is no immediate job to dispatch, and workers are entering wait state
- make `setup_connection()` recreate the `MYSQL` handle on demand when a previously released pooled slot is borrowed again

## Local verification
- verified a clean precondition before restore: no leftover `myloader` test processes and the target verification database was empty (`BASE TABLE` / `VIEW` / `TRIGGER` / `ROUTINE` / `EVENT` all zero)
- ran a selective restore of 10 large tables from a local dump directory
- used `-t 20 --max-threads-per-table 20`
- baseline `PROCESSLIST` at start was 2 total application sessions
- shortly after start it rose to 22 total sessions
- near the tail it dropped to 7 total sessions while restore was still running, confirming that idle pooled connections were being released

## Notes
This does not address the separate end-of-run behavior where interrupted restores can still leave long-running `ALTER TABLE` work after timeout/kill. That remains separate from this PR.

### DIFF
--- a/src/myloader/myloader_restore.c
+++ b/src/myloader/myloader_restore.c
@@ -187,12 +187,30 @@ void wait_restore_threads_to_close(){
 }
 
 void reconnect_connection_data(struct connection_data *cd){
-  mysql_close(cd->thrconn);
+  if (cd->thrconn != NULL)
+    mysql_close(cd->thrconn);
   cd->thrconn=mysql_init(NULL);
   m_connect(cd->thrconn);
   cd->connection_id=mysql_thread_id(cd->thrconn);
   execute_use(cd);
   execute_gstring(cd->thrconn, set_session);
+}
+
+gboolean release_idle_connection_if_possible(){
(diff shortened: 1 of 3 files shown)


davidducos (approved):
Looks good to me

PR #2271, opened 2026-05-08
Fixing break in ignore_engines Check

jihan-lf:
Caused abandonment of table enumeration for the entire database on the first match of an exclusion rule


jihan-lf:
This was discovered in a mixed-engine database (InnoDB + Columnstore)

PR #2276, opened 2026-05-14
Implement --only-indexes flag

artm:
Implement `--only-indexes` to skip schema and data and only load secondary indexes.

Example use case: loading a database with large indexes on a shared galera cluster. To avoid the cluster locking up on large `ALTER TABLE` statements with `wsrep_OSU_method = TOI` we would like to execute these specific statements with `wsrep_OSU_method = NBO`. But the rest of the load process is incompatible with this mode, so we would split the loading into two phases:

1. `--drop-database --skip-indexes` : recreates the database, loads schema except indexes and data
2. `--only-indexes` with `wsrep_OSU_method = NBO` set via custom `--defaults-extra-file` : leaves database, schema and data alone, assuming they already exist, create indexes.

See https://github.com/mydumper/mydumper/discussions/2272 for initial discussion.


davidducos:
Hey @artm, 
The --optimize-keys option is use and could be expended for the way how secondary indexes are added, it is not the intention to filter in or filter out. 

On https://github.com/mydumper/mydumper/pull/2275, we are changing 3 things
1- Every single table will be split, before depended on some rules.
2- We are able to control with the --skip options what is not going to be imported, allowing us all the possible combinations.
3- We kept --no-schemas which should be used at a higher level.

I think that you will be able to achieve the same goal doing:

--drop-database --skip-indexes --skip-constraints
--skip-create-table with wsrep_OSU_method = NBO set via custom --defaults-extra-file

Can you review #2275 ? as I will be merging it later today.

artm:
Hi @davidducos 

I have misinterpreted `--optimize-keys=SKIP` this morning. I did understand it correctly yesterday while testing various scenarios but forgot what I understood when attempting my solution.

1. `--drop-database --skip-indexes --skip-constraints` - looks good, will recreate the database, load schema, load data but not indexes/constraints, so will be faster. This is exactly what I wanted to achieve in my proposal.

2. `--skip-create-table ` + `wsrep_OSU_method = NBO` will then have to be combined with `--skip-data` because we have loaded data already and will not create indexes because with `no_data` the index threads will be shut down early.

It is still possible to achieve what we want this way:

1. `--drop-database --skip-indexes --skip-constraints --skip-data` : recreate the database, load schema, don't load data.

2. `--skip-create-table ` + `wsrep_OSU_method = NBO` : load data, then create secondary indexes.

I think that will work.

Could we add a testcase similar to the one in this pull request to verify that it will work as intended? Would it be possible to review which DDL statements myloader executes when ran with `--skip-create-table ` ?

artm:
Hi again @davidducos 

If I understand correctly, even with the new `--skip-create-table`, `CREATE DATABASE IF NOT EXISTS` will be executed which is not compatible with `wsrep_OSU_method = NBO` at least on Percona Xtradb Cluster 8.4.8. So the phase 2 run in the above comments is still not possible. 

In my branch it will not be executed because I made the (incorrectly chosen) `--optimize-keys=ONLY` to imply no data and **no schema**.

davidducos:
Hi @artm , 
There is no parameter like `--skip-create-database`, we could add it. As a workaround, If you want to avoid the execution of a `CREATE DATABASE` statement, you can delete the `<database>-schema-create.sql` file from the backup, if the database is not there it will fail like this:
```
root@e2bb45b64c51:~/issue_2272/mydumper# rm data/sakila-schema-create.sql
root@e2bb45b64c51:~/issue_2272/mydumper# mysql -e "drop schema sakila;"
root@e2bb45b64c51:~/issue_2272/mydumper# ./myloader --verbose 4 --checksum-all -d data -o --max-threads-for-schema-creation=1
** Message: 12:32:58.770: MyDumper restore version: 1.0.0-1
** Message: 12:32:58.770: Using 4 loader threads
...
** Message: 12:32:58.813: L-Thread 1: Starting import
** Message: 12:32:58.813: L-Thread 2: Starting import
** Message: 12:32:58.813: S-Thread 5: Starting import
** Message: 12:32:58.813: L-Thread 3: Starting import
** Message: 12:32:58.813: Thread 5: restoring table sakila.staff_list from /home/circleci/project/issue_2272/mydumper/data/sakila.staff_list-schema.sql
** Message: 12:32:58.813: Dropping table or view (if exists) sakila.staff_list
** Message: 12:32:58.813: L-Thread 4: Starting import

** (myloader:23076): WARNING **: 12:32:58.871: Thread 5: Error switching to database `sakila` - ERROR 1049: Unknown database 'sakila'

** (myloader:23076): CRITICAL **: 12:32:58.871: Thread 5 with connection 1366: Error switching to database `sakila` request_another_connection: Unknown database 'sakila'
```
I hope this helps.

artm:
Hi @davidducos 

That sounds like it would work. In order to test we would need a version of `mydumper` with both `--skip-create-table` and `--skip-create-database`, or suggested workaround for the second. Do you have an estimate when will `mydumper` (pre)release with at least `--skip-create-table` be available? We would like to avoid having to build it ourselves. Not because it is difficult, but because that would complicate our own internal processes.

davidducos:
hi @artm , I will be releasing v1.0.1 later today or tomorrow. About --only-index, is similar to use the --skip on the other objects. I think that this could be consider as a shortcut, like `--only-index  == --skip-create-table && --skip-constraints` and honestly, I did that in the code and now I regret, as for instance --no-schema is a similar example, I prefer the user specify what they want and what they don't want... but on the other hand it is not the same, as --no-schema is a global setting, and only-index is a shortcut, actually if we could implement --no-schema like a shortcut will be even better.
BTW, https://github.com/mydumper/mydumper/pull/2277 will be merge soon which implemented the --skip-create-database.

PR #2278, opened 2026-05-18
fix: correct current_threads_running accounting in --use-defer path (livelock)

prgTW:
## Problem

When `--use-defer` is active and a database contains any table with an integer primary key, mydumper hangs indefinitely after printing row-count estimates. No data is ever written. The process must be killed manually.

### Root cause

The livelock lives in the interaction between the chunk builder thread (`table_job_enqueue` in `mydumper_chunks.c`) and the worker threads (`process_job` in `mydumper_working_thread.c`).

```mermaid
sequenceDiagram
    participant CB as Chunk Builder
    participant QQ as q->queue
    participant QD as q->defer
    participant W  as Worker Thread

    Note over CB,W: --use-defer active · integer-PK table · max_threads_per_table = 4

    CB->>CB: get_next_dbt_and_chunk_step_item()<br/>dbt->current_threads_running++ → 1
    CB->>QD: push JOB_DUMP  (real work)
    CB->>QQ: push JOB_DEFER (sentinel)

    W-->>QQ: pop JOB_DEFER
    Note over W: case JOB_DEFER: break<br/>❌ counter NOT decremented<br/>current_threads_running stays 1

    CB->>CB: get_next_dbt_and_chunk_step_item()<br/>dbt->current_threads_running++ → 2
    CB->>QD: push JOB_DUMP
    CB->>QQ: push JOB_DEFER

    W-->>QQ: pop JOB_DEFER
    Note over W: case JOB_DEFER: break<br/>❌ counter stays 2

    Note over CB: … repeat until current_threads_running == max_threads_per_table …

    loop Livelock — spins forever
        CB->>CB: get_next_dbt_and_chunk_step_item()<br/>current_threads_running (4) ≥ max (4) → returns NULL
        CB->>CB: usleep(1) · retry
    end

    Note over QQ,QD: q->queue never receives JOB_SHUTDOWN<br/>workers never reach process_queue(q->defer, …)<br/>real JOB_DUMPs sit in q->defer forever
```

**The invariant that breaks:** `dbt->current_threads_running` is incremented in `get_next_dbt_and_chunk_step_item()` for every chunk routed to `q->defer`, but it is only decremented inside `thd_JOB_DUMP()` — which only runs when a real `JOB_DUMP` is consumed from `q->queue`. Because the matching `JOB_DEFER` sentinel is a bare `break`, the counter grows monotonically until it equals `max_threads_per_table`, after which the chunk builder sees every table as "full" and spin-loops on `usleep(1)`. `q->queue` never receives `JOB_SHUTDOWN`, so workers never advance to `process_queue(q->defer, …)` where the real work is waiting.

The bug affects **any** database that contains at least one integer-PK table — not only all-integer-PK databases.

## Fix

Add the symmetric decrement to the `JOB_DEFER` handler:

```mermaid
sequenceDiagram
    participant CB as Chunk Builder
    participant QQ as q->queue
    participant QD as q->defer
    participant W  as Worker Thread

    Note over CB,W: --use-defer active · integer-PK table · max_threads_per_table = 4

    CB->>CB: get_next_dbt_and_chunk_step_item()<br/>dbt->current_threads_running++ → 1
    CB->>QD: push JOB_DUMP
    CB->>QQ: push JOB_DEFER

    W-->>QQ: pop JOB_DEFER
    Note over W: case JOB_DEFER:<br/>✅ dbt->current_threads_running-- → 0

    CB->>CB: get_next_dbt_and_chunk_step_item()<br/>current_threads_running (0) < max (4) → OK
    CB->>QD: push JOB_DUMP
    CB->>QQ: push JOB_DEFER

    Note over CB: … all chunks enqueued, table list drains to empty …

    CB->>QQ: JOB_SHUTDOWN (enqueue_shutdown)
    W-->>QQ: pop JOB_SHUTDOWN · exit phase 1
    W-->>QD: process_queue(q->defer, …) · drain real JOB_DUMPs ✅
```

The chunk builder's existing `max_threads_per_table` spin-loop already pushes back to `request_chunk` on each pass, so no additional signal is required — the guard clears naturally after the decrement and the builder proceeds.

## Commits

This PR contains two commits following a TDD approach:

1. **test: prove --use-defer livelock** — adds a self-contained GLib-only unit test (`test/unit/test_use_defer_livelock.c`) that models the counter state machine and asserts the post-fix contract. The test exits non-zero on the unfixed code, proving the bug. No live database required.

2. **fix: correct current_threads_running accounting** — adds the symmetric decrement to the `JOB_DEFER` handler and enables `-DUSE_DEFER_FIX` in the CMake target, making the test green.

To run the tests:
```sh
cmake -B build . && cmake --build build
ctest --test-dir build --output-on-failure
```

## Why a unit test rather than an integration test

The livelock's failure mode is a **hang**, not a non-zero exit code. An integration test running mydumper against a real database would block indefinitely on the unfixed code — there is no SQL error to catch, just a thread spinning forever. Making that fail fast on CI would require wrapping the test runner in a shell `timeout`, which adds fragility and obscures what is actually being tested.

The bug is entirely in the counter state machine: one increment in `get_next_dbt_and_chunk_step_item()`, one missing decrement in `process_job()`. A unit test that models those two operations directly proves the invariant in under a second, compiles with only GLib (no database, no network), and runs on every CI environment in the matrix.

### DIFF
--- a/.circleci/circleci_config_builder.sh
+++ b/.circleci/circleci_config_builder.sh
@@ -480,6 +480,9 @@ cat <<EOF
           source /etc/profile.d/sh.local || true
     - run: cmake . <<parameters.CMAKED>>
     - run: make VERBOSE=1
+    - run:
+        name: Run unit tests
+        command: ctest --output-on-failure
     - run: sudo make install
     - run: ./mydumper --version
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2494,6 +2494,9 @@ commands:
           source /etc/profile.d/sh.local || true
     - run: cmake . <<parameters.CMAKED>>
     - run: make VERBOSE=1
+    - run:
+        name: Run unit tests
+        command: ctest --output-on-failure
     - run: sudo make install
     - run: ./mydumper --version
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -120,6 +120,16 @@ add_custom_target(dist
 
 OPTION(RUN_CPPCHECK "Run cppcheck" OFF)
 
+enable_testing()
+
+# Unit test: --use-defer livelock
+# Requires only GLib (no MySQL). Link with -DUSE_DEFER_FIX to test the fixed
+# code path; without it the test proves the bug (exits non-zero).
+add_executable(test_use_defer_livelock test/unit/test_use_defer_livelock.c)
+target_include_directories(test_use_defer_livelock PRIVATE ${GLIB2_INCLUDE_DIR})
+target_link_libraries(test_use_defer_livelock ${GLIB2_LIBRARIES})
+add_test(NAME use_defer_livelock COMMAND test_use_defer_livelock)
+
 IF(RUN_CPPCHECK)
   include(CppcheckTargets)
   add_cppcheck(mydumper)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -128,6 +128,8 @@ enable_testing()
 add_executable(test_use_defer_livelock test/unit/test_use_defer_livelock.c)
 target_include_directories(test_use_defer_livelock PRIVATE ${GLIB2_INCLUDE_DIR})
 target_link_libraries(test_use_defer_livelock ${GLIB2_LIBRARIES})
+# -DUSE_DEFER_FIX tells the test to exercise the fixed JOB_DEFER handler path.
+target_compile_definitions(test_use_defer_livelock PRIVATE USE_DEFER_FIX)
 add_test(NAME use_defer_livelock COMMAND test_use_defer_livelock)
 
 IF(RUN_CPPCHECK)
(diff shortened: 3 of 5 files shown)


prgTW:
Hey @davidducos 👋. Some time ago I encountered `mydumper` hanging forever when using the `--defer` option and came up with this fix. First commit to prove the situation, second - the fix. I also included charts to better describe the situation, but wanted to ask You to take a look and check if this fix makes sense and follows Your code standards.

I don't want to lie to You - it was AI-coded after examination of the problem and I'm not specifically a C programmer. Just wanted to help 🙂

davidducos:
Hi @prgTW , thank you for finding that bug! I usually do not user --defer, so I was not going to be able to get it. I will start using AI more often on MyDumper from now on, so, no worries about it. 
The code is excellent! I changed just one thing related to how we build that config.yml file, which we might be changing on the future, so, no worries anyways, but for now we have to be consistent with the configuration. Thanks again!! I will be releasing v1.0.3-1 this week due another bug and I will be packing this too.

prgTW:
Great, thanks for the additional commit! I'm already looking forward to using this version, thank You for merging :)

PR #2280, opened 2026-05-24
Fixing SAFE_NO_LOCK bug

davidducos:
This is major fix over the SAFE_NO_LOCK mode of the parameter --sync-thread-lock-mode. 
The problem was that the get_binlog function was executed inside a thread that might or might not be at the same point in time of the other threads. 
The logic was move to the main connection which now will execute both, the initial point in time and after all the threads are in ready. 
Again, SAFE_NO_LOCK considers that if there is no change in the binlog between the time all the threads send the START TRANSACTION WITH CONSISTENT SNAPSHOT, the backup will be consistent.

### DIFF
--- a/src/mydumper/mydumper_start_dump.c
+++ b/src/mydumper/mydumper_start_dump.c
@@ -1009,6 +1009,31 @@ cleanup:
     mysql_free_result(slave);
 }
 
+static
+void get_binlog_position(MYSQL *conn, char **masterlog, char **masterpos, char **mastergtid){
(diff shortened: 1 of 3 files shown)


idan3:
@davidducos Are pre 1.*.* versions affected? (0.21.* ?)
If I do not set `--sync-thread-lock-mode` at all does it affect me?

davidducos:
Hi @idan3 , this affected all version previous v1.0.3 that implemented SAFE_NO_LOCK

PR #2304, opened 2026-07-13
Add AWS session parameter support

daniel-percona:
Allows myloader to disable binlogging on Aurora/MySQL 5.7 using AWS session calls, and supports multiple AWS-specific session commands per connection.


daniel-percona (commented):
Modified example config with explanation/use case

davidducos:
Hi @daniel-percona !! thank you for your PR, but I'm wondering why did you add a new file myloader.cnf as you can use the /etc/mydumper.cnf file to load the myloader config ? if it is just to keep record of the possible and expecific configuration for AWS, we can add them commented in the mydumper.cnf.
Please, lef me know your thoughts about it and if you want me to move them.

daniel-percona:
> Hi @daniel-percona !! thank you for your PR, but I'm wondering why did you add a new file myloader.cnf as you can use the /etc/mydumper.cnf file to load the myloader config ? if it is just to keep record of the possible and expecific configuration for AWS, we can add them commented in the mydumper.cnf. Please, lef me know your thoughts about it and if you want me to move them.

Hey Dave, no worries, that was just to separate it in case we need to use different mysql specific config sections for the source and destination hosts, but we can keep it together in the repo as you had originally.

PR #2309, opened 2026-07-16
Fix segfault when EXPLAIN output has no rows column (explain_format=TREE/JSON)

### LINKED ISSUE #2307: [BUG] Segmentation fault at start of data export against MySQL 9.7.1 (v1.0.3-1, statically-linked MySQL 8.4.9 client)
**Describe the bug**

`mydumper` v1.0.3-1 segfaults reliably at the very start of the **data export** phase when dumping from a **MySQL 9.7.1** server. The schema phase completes for all tables; the crash happens immediately after `Starting exporting data for Transactional tables`, before any row data is written. It reproduces even with a single 2-row table and cannot be worked around.

**Environment**

- mydumper: `v1.0.3-1, built against MySQL 8.4.9 with SSL support` (official `mydumper_1.0.3-1.noble_amd64.deb` from the GitHub release)
- MySQL client: **statically linked** into the binary — `ldd $(command -v mydumper)` shows only `libssl.so.3`, no `libmysqlclient`/`libmariadb`
- Server: `MySQL 9.7.1 Community Server - GPL`
- OS: Ubuntu 26.04 LTS, x86_64

**To Reproduce**

Even the most minimal command crashes — a single small table (2 rows), no compression, default threads:

```
mydumper --database db_name --tables-list db_name.ar_internal_metadata --outputdir /tmp/md-one --verbose 3
```

**Log (single-table repro, trimmed tail)**

```
** Message: MyDumper backup version: 1.0.3-1
** Message: Connected to MySQL 9.7.1
...
** Message: Thread 2: dumping schema for `db_name`.`ar_internal_metadata`
** Message: Thread 3: Schema jobs are done, Starting exporting data for Non-Transactional tables
** Message: Enqueuing of non-transactional tables completed
** Message: Starting to enqueue transactional tables
** Message: Thread 1: Non-Transactional tables are done, Starting exporting data for Transactional tables
Segmentation fault   (exit code 139)
```

The crash always happens at the transition into the data-export phase, before any rows are dumped.

**Kernel log (dmesg)** — the fault is inside the `mydumper` binary (statically-linked client), deterministic offset, read fault:

```
data[249418]: segfault at 7b86443ef900 ip ...90d error 4 in mydumper[7e90d,...+188000]
```

The faulting instruction offset (`...90d`) is identical on every run; the faulting thread is a data worker (renamed `data`).

**Isolation performed** (all still segfault unless noted)

- Full production command (`--source-data --replica-data --compress --triggers --events --routines --rows 1000000 --trx-tables`) → segfault
- Minimal command, no replication flags → segfault
- Without `--compress` → segfault
- Single 2-row table (`ar_internal_metadata`), single thread → segfault
- `--server-version "MySQL - 9.7.1"` → segfault
- `--server-version "MySQL - 8.4.0"` → segfault

So the crash is independent of compression, table size/content, replication-coordinate options, thread count, and the `--server-version` override. **Schema export works; data export crashes immediately.**

**Expected behavior**

Data export completes against MySQL 9.7.x — or at least fails with a clear error instead of `SIGSEGV`.

**Note**

The binary statically links the MySQL **8.4.9** client while the server is **9.7.1**. This looks like it could be an incompatibility between the bundled 8.4.9 client and a 9.7.1 server in the row/result-set fetch path (or a mydumper bug in that path when talking to 9.7). I can't install `gdb` on this host, but I'm happy to run other non-invasive diagnostics if that helps narrow it down.

pioz:
## Problem

`mydumper` crashes with SIGSEGV at the very start of the data export phase when the server returns `EXPLAIN` output without a `rows` column. This happens on MySQL 8.3+ when the `explain_format` system variable is set to `TREE` or `JSON`: a plain `EXPLAIN SELECT ...` then returns a single column named `EXPLAIN`, with no `rows` column.

The culprit is in `get_rows_from_explain()`:

```c
guint row_col=-1;                             // guint, so 4294967295
determine_explain_columns(mr->res, &row_col); // sets row_col only if a rows/estRows column exists
if ( mr->row[row_col]==NULL){                 // wild out-of-bounds read -> SIGSEGV
```

When `determine_explain_columns()` does not find a `rows`/`estRows` column, `row_col` keeps its initial value `(guint)-1` and `mr->row[4294967295]` reads arbitrary memory. This matches all the symptoms reported in #2307: deterministic crash right after "Starting exporting data for Transactional tables" (before any "has ~N rows" message is printed), identical faulting instruction offset on every run, and no effect from `--compress`, thread count or `--server-version` (the crash depends on the shape of the server's EXPLAIN output, not on the parsed version).

## Reproduction

Against MySQL 8.4 started with `--explain-format=TREE`, any dump crashes at the start of the data export phase, even for a single 2-row table:

```
mydumper --database db_name --tables-list db_name.some_table --outputdir /tmp/out
```

## Fix

- `determine_explain_columns()` now returns whether the `rows`/`estRows` column was found.
- When the column is missing, `get_rows_from_explain()` retries once forcing `EXPLAIN FORMAT=TRADITIONAL` (only for MySQL-family servers; MariaDB and TiDB always include the column, so they never reach the retry).
- If the column is still missing, it logs a warning and returns 0 rows (falling back to a full table scan) instead of crashing.

## Verification

Tested against MySQL 8.4.10 started with `--explain-format=TREE` and a 2-row table:

- unpatched build: deterministic segfault, log identical to #2307;
- patched build: the retry kicks in, the row count is estimated correctly ("has ~2 rows") and the dump completes with the correct data;
- with `explain_format=TRADITIONAL` the behavior is unchanged (the retry never triggers).

I have also been running this patch live against the server where the issue was originally reported and the dump now completes successfully.

Fixes #2307

I'm not sure this is fully ready to merge as-is, but hopefully it can be a good starting point for the fix.

This PR was opened with the help of Claude Code.

### DIFF
--- a/src/mydumper/mydumper_chunks.c
+++ b/src/mydumper/mydumper_chunks.c
@@ -256,16 +256,40 @@ cleanup:
 }
 
 
+// Only the MySQL family supports the explain_format system variable (added in
+// MySQL 8.3), which can change the output of a plain EXPLAIN to TREE or JSON,
+// and accepts forcing the format back with EXPLAIN FORMAT=TRADITIONAL
+static
+gboolean explain_format_can_be_forced(){
+  switch (get_product()){
+    case SERVER_TYPE_MYSQL:
+    case SERVER_TYPE_PERCONA:
+    case SERVER_TYPE_RDS:
+    case SERVER_TYPE_GOOGLE:
+    case SERVER_TYPE_UNKNOWN:
+      return TRUE;
+    default:
+      return FALSE;
+  }
+}
+
 guint64 get_rows_from_explain(MYSQL * conn, struct db_table *dbt, GString *where, gchar *field){
-  gchar *query = g_strdup_printf(
-                        "EXPLAIN SELECT %s %s%s%s FROM %s%s%s.%s%s%s%s%s",
+  const gchar *explain_statement = "EXPLAIN";
+  gchar *query = NULL;
+  struct M_ROW *mr = NULL;
+  guint row_col = 0;
+
+execute_explain:
+  query = g_strdup_printf(
+                        "%s SELECT %s %s%s%s FROM %s%s%s.%s%s%s%s%s",
+                        explain_statement,
                         is_mysql_like() ? "/*!40001 SQL_NO_CACHE */": "",
                         field?identifier_quote_character_str:"", field?field:"*", field?identifier_quote_character_str:"",
                         identifier_quote_character_str, dbt->database->source_database, identifier_quote_character_str, identifier_quote_character_str, dbt->table, identifier_quote_character_str,
                         where?" WHERE ":"",where?where->str:"");
   /* Get minimum/maximum */
   trace("EXPLAIN: %s", query);
-  struct M_ROW *mr = m_store_result_row(conn, query, 
+  mr = m_store_result_row(conn, query,
                         m_critical, m_warning, "Failed to execute EXPLAIN: %s", query);
 
   g_free(query);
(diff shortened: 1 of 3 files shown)


davidducos:
Hi @pioz , 
I'm getting:
image
when I try to resolve the conflicts. Can you grant me access or resolve the conflicts?

pioz:
> Hi @pioz , I'm getting: image when I try to resolve the conflicts. Can you grant me access or resolve the conflicts?

Done! ✅

davidducos (approved):

PR #2311, opened 2026-07-24
Code Formatting Standard

utdrmac:
This PR introduces `clang-format`, along with a basic set of formatting rules, based on the Goggle C/C++ [style guidelines](https://google.github.io/styleguide/cppguide.html).

There are no logic changes, or features, or bug fixes in this PR. This PR is solely a code format change.

Also introduced, is a github workflow to enforce the code format. On any PR that modifies .c or .h files inside src/, the workflow will execute to verify the files adhere to the format. The workflow **does not** modify/reformat, it only validates. If vaildation fails, the PR OP must execute `clang-format` in their branch, and then push the updated code.


utdrmac:
Commit dcd7292 has a bit of code reorganization to fix a circular include dependency issue.

PR #2313, opened 2026-07-27
Add EXPLAIN-based string primary-key chunk planner

daniel-percona:
This PR introduces a metadata-assisted planner for tables with string primary keys, especially large tables where the previous chunk-discovery process could take hours.

The new planner uses EXPLAIN statements to estimate prefix ranges instead of executing data-reading statements such as:

```
SELECT DISTINCT LEFT(product_ari, 1) ...
SELECT LEFT(product_ari, 1) ... LIMIT 1
```

Those queries could scan a very large portion of the primary-key index before export began. The new approach keeps discovery read-only and uses optimizer estimates without reading the table data.

The data-export queries still execute normally after planning. Only the initial range-discovery process was changed.

### New Features

- Adds --string-pk-planner=auto|metadata|recursive.
- Adds configurable planner timeout and probe limits.
- Adds configurable maximum root prefixes.
- Adds --string-pk-planner-max-depth.
- Uses EXPLAIN-only prefix discovery.
- Uses bounded first-level discovery by default.
- Allows normal chunk workers to refine the discovered roots.
- Falls back safely to the existing recursive splitter when required.

### Correctness Fixes

- Prevents overlapping LIKE ranges.
- Escapes %, _, and \ correctly.
- Handles case-insensitive collations.
- Preserves complete composite primary keys.
- Corrects half-open string range boundaries.
- Prevents duplicate rows from being exported into multiple chunks.


daniel-percona:
I tested PR #2304 and PR #2313 together from origin/master.

The branches were merged locally, and the conflicts in CMakeLists.txt
and README.md were resolved while preserving both features.

The combined build passed all four tests. PR #2313 depends on PR #2304
being merged first.

PR #2330, opened 2026-08-20
Warn when --regex excludes the database object (#2329)

### LINKED ISSUE #2329: [BUG] --regex '^db\.' silently omits db-schema-create.sql; in --stream this defers the whole restore to EOF
<html>
<body>
<html><head></head><body><p><strong>Describe the bug</strong></p>
<p><code>--regex</code> is matched against two different subjects: <strong><code>database.table</code></strong> for tables, but the <strong>bare
database name</strong> for the database object itself.</p>
<p>So <code>--regex '^mydb\.'</code> selects every table in <code>mydb</code> and silently drops <code>mydb</code> itself —
<code>mydb-schema-create.sql</code> is never written. <strong>mydumper still exits 0</strong>, so the dump looks complete.</p>
<p>Two consequences, both measured below:</p>
<ol>
<li><strong>If the target database does not exist, the dump cannot be restored</strong> — <code>ERROR 1049</code>, 0 tables, 0 rows.</li>
<li><strong>If it does exist, the restore succeeds but <code>--stream</code> stops pipelining.</strong> Every schema job is held
until the stream reaches EOF, so myloader creates nothing and loads nothing for the whole duration
of the dump.</li>
</ol>
<p>(2) is the dangerous one: no error, no warning, no progress. On a 110 GB production migration it looked
like a hung loader for hours, and killing mydumper appeared to "fix" it — that was just EOF releasing
the held jobs.</p>
<p>The fix on the user side is one character:</p>
<pre><code class="language-diff">- --regex '^(db1\.|db2\.|db3\.)'     # tables only; the database object is dropped
+ --regex '^(db1|db2|db3)(\.|$)'     # correct
</code></pre>
<hr>
<p><strong>To Reproduce</strong></p>
<p>Two MySQL 8.0 containers plus <code>mydumper/mydumper:v1.0.3-1</code>. <code>-B</code> / <code>--database</code> is never passed to
myloader. Each case starts from a fresh, non-existent <code>-o</code> / <code>-d</code> path.</p>
<p>Bad pattern, non-stream:</p>
<pre><code class="language-bash">mydumper -h $SRC -P 3306 -u root -p $PW --regex '^repro_db\.' \
  -o /tmp/dump_a -v 3 -L /artifacts/dump_a.log

myloader  -h $TGT -P 3306 -u root -p $PW \
  -d /tmp/dump_a -v 3 -L /artifacts/load_a.log --debug
</code></pre>
<p>Bad pattern, stream:</p>
<pre><code class="language-bash">mydumper -h $SRC -P 3306 -u root -p $PW --regex '^repro_db\.' \
  -o /tmp/dump_b --stream -v 3 -L /artifacts/dump_b.log \
| myloader -h $TGT -P 3306 -u root -p $PW \
  -d /tmp/load_b --stream -v 3 -L /artifacts/load_b.log --debug
</code></pre>
<p>Corrected pattern, stream (control):</p>
<pre><code class="language-bash">mydumper -h $SRC -P 3306 -u root -p $PW --regex '^repro_db(\.|$)' \
  -o /tmp/dump_c --stream -v 3 -L /artifacts/dump_c.log \
| myloader -h $TGT -P 3306 -u root -p $PW \
  -d /tmp/load_c --stream -v 3 -L /artifacts/load_c.log --debug
</code></pre>
<p>Version used:</p>
<pre><code>mydumper v1.0.3-1, built against MySQL 8.4.9 with SSL support
myloader v1.0.3-1, built against MySQL 8.4.9 with SSL support
</code></pre>
<p>Result, with the target database dropped or pre-created before each run:</p>

  | target DB absent | target DB present (empty)
-- | -- | --
non-stream | exit 1 after ~1 s — ERROR 1049: Unknown database, 0 tables | exit 0, 5/5 tables, 4,000,000 rows
--stream | exit 1 at stream EOF (103 s) — same error, after all 851 MB had been transferred and discarded | exit 0 (163 s), 4,000,000 rows — but nothing created or loaded until EOF

<p>In the last row the five schema jobs enter the queue at ~30 s and sit untouched for 63 s. On a 110 GB
dump that dead window is hours, with 0 tables, 0 rows, no error and no warning.</p>
<p>Without <code>--debug</code>, the only thing the loader prints during that window is:</p>
<pre><code>[INFO] - Reading metadata: metadata.partial.N
[INFO] - Removing file: &lt;dir&gt;/metadata.partial.N
</code></pre>
<p>Full logs can be attached on request.</p>
<hr>
<p><strong>Backup</strong></p>
<p>Bad pattern — 5 table-schema files, 8 data files, <code>metadata</code>, and <strong>no</strong> <code>repro_db-schema-create.sql</code>:</p>
<pre><code>$ ls -l /tmp/dump_a
-rw-r--r--        445  repro_db.t1-schema.sql
-rw-r--r--  170308880  repro_db.t1.00000.sql
-rw-r--r--        445  repro_db.t2-schema.sql
-rw-r--r--  158932585  repro_db.t2.00000.sql
-rw-r--r--   11361130  repro_db.t2.00001.sql
-rw-r--r--        445  repro_db.t3-schema.sql
-rw-r--r--   70890424  repro_db.t3.00000.sql
   ... t3.00001 .. t3.00003 ...
-rw-r--r--        445  repro_db.t4-schema.sql
-rw-r--r--  170293393  repro_db.t4.00000.sql
-rw-r--r--        445  repro_db.t5-schema.sql
-rw-r--r--  170293441  repro_db.t5.00000.sql
-rw-r--r--       1019  metadata

$ ls /tmp/dump_a | grep schema-create
(no match)
</code></pre>
<p>The difference between the two patterns, in one line each:</p>
<pre><code>bad      : grep -oE 'repro_db[^ ]*\.sql' dump_b.log | sort -u  -&gt;  10 files, no *-schema-create.sql
corrected: grep -oE 'repro_db[^ ]*\.sql' dump_c.log | sort -u  -&gt;  11 files, incl. repro_db-schema-create.sql
</code></pre>
<hr>
<p><strong>How to repeat</strong></p>
<p>Any table works; the only thing that matters is that the dump runs long enough for the stall to be
visible. I used 5 tables × 800,000 rows (~813 MB) and capped the tool container to <code>--cpus 0.15</code>, which
gave a 37 s dump and a 91–103 s stream.</p>
<pre><code class="language-sql">CREATE DATABASE repro_db;
USE repro_db;
SET SESSION cte_max_recursion_depth = 200000;

CREATE TABLE t1 (id INT NOT NULL AUTO_INCREMENT PRIMARY KEY,
                 pad CHAR(200) NOT NULL) ENGINE=InnoDB;

INSERT INTO t1 (pad)
  WITH RECURSIVE s(n) AS (SELECT 1 UNION ALL SELECT n+1 FROM s WHERE n &lt; 50000)
  SELECT RPAD(CONCAT('row-', n, '-'), 200, 'x') FROM s;
INSERT INTO t1 (pad) SELECT pad FROM t1;   -- repeat 4x -&gt; 800,000 rows

CREATE TABLE t2 LIKE t1;  INSERT INTO t2 (pad) SELECT pad FROM t1;
CREATE TABLE t3 LIKE t1;  INSERT INTO t3 (pad) SELECT pad FROM t1;
CREATE TABLE t4 LIKE t1;  INSERT INTO t4 (pad) SELECT pad FROM t1;
CREATE TABLE t5 LIKE t1;  INSERT INTO t5 (pad) SELECT pad FROM t1;
</code></pre>
<p>Then run the three commands above. On the target, while the stream case runs:</p>
<pre><code class="language-sql">SELECT COUNT(*) FROM information_schema.tables WHERE table_schema='repro_db';
</code></pre>
<p>It stays at <code>0</code> for the entire dump and jumps to <code>5</code> in the same sample where mydumper exits.</p>
<hr>
<p><strong>Environment (please complete the following information):</strong></p>
<ul>
<li>OS version: Ubuntu 22.04 (production, 2 vCPU) / AlmaLinux 9.8 inside the reproduction image</li>
<li>MyDumper version: <code>mydumper v1.0.3-1, built against MySQL 8.4.9 with SSL support</code></li>
<li>MySQL: reproduction 8.0.46 both ends; production Azure Database for MySQL Flexible Server
8.0.46 → 8.0.45</li>
<li>Reproduction image: <code>mydumper/mydumper:v1.0.3-1</code>, linux/arm64</li>
</ul>
<hr>
<p><strong>Additional context</strong></p>
<p><em>How this presented in the field, in case it helps someone searching.</em> On a 3-database / ~326-table /
~110 GB migration the loader simply did nothing for hours while data kept arriving. I ruled out, about
an hour each and with nothing in the log to go on: a missing <code>--rows</code>; too many schemas at once; the
external <code>zstd</code> decompressor; and schema-thread autoscale on a 2-vCPU host. Only <code>--debug</code> plus reading
<code>schema_push()</code> showed that the database object had never been created — because the regex required a
dot. One character, one day.</p>
<p><em>Suggested fix — warn at dump time (mydumper).</em> This catches the problem before any time is wasted,
regardless of how the restore later fails. Once per database, when tables from it are selected but the
database object was excluded by <code>--regex</code>:</p>
<pre><code>WARNING: --regex selected tables from `mydb` but not the database object itself, so
         `mydb-schema-create.sql` will not be dumped. Database objects are matched
         against the bare database name — use e.g. '^(mydb)(\.|$)'.
</code></pre>
<p><em>Suggested fix — warn at restore time (myloader).</em> Once per database, where schema jobs start being held:</p>
<pre><code class="language-diff">  /* myloader_worker_schema.c — schema_push(), already under _database-&gt;mutex */
    if (schema_worker_job == SCHEMA_TABLE_JOB ) {
+     if (!_database-&gt;create_missing_warned) {
+       _database-&gt;create_missing_warned = TRUE;
+       g_warning("No CREATE DATABASE received for `%s`; schema jobs will be held until "
+                 "the end of the stream. If you are using --regex, database objects are "
+                 "matched against the bare database name.", _database-&gt;target_database);
+     }
      g_async_queue_push(_database-&gt;table_queue, sj);
</code></pre>
<p><em>Suggested fix — docs.</em> State the two-subject rule in the <code>--regex</code> help text, with
<code>'^(db1|db2)(\.|$)'</code> as the example.</p>
<p>I'd be glad to open a PR for the mydumper-side warning together with a <code>test/specific_NN/</code> regression
case, and follow up with the myloader one.</p>
<p><em>Question for maintainers.</em> Would it be reasonable for myloader to fall back to
<code>CREATE DATABASE IF NOT EXISTS</code> when no schema-create file arrives, instead of holding every schema job
until EOF? <code>create_database()</code> already does this on the <code>-B</code> / <code>--database</code> path. That would remove the
stream stall entirely, but it is a behaviour change so I did not assume it.</p>
<p><em>Two notes on the issue template itself.</em> <code>--verbose 4</code> is not accepted — <code>common_options.c:196</code>
documents <code>0 = silent, 1 = errors, 2 = warnings, 3 = info</code>. And <code>--debug</code> on its own prints nothing:
<code>trace()</code> resolves to <code>g_debug()</code> (<code>src/common.c:1366</code>), which GLib discards unless <code>G_MESSAGES_DEBUG</code>
is set. With <code>G_MESSAGES_DEBUG=all</code> the same run went from 27 to 401 log lines. Both might be worth
correcting in the template and the <code>--debug</code> help text.</p>
<p><em>Related.</em> Possibly related to #1899, though the root cause may differ — there, tables <em>were</em> being
created (339 of 680), which implies the database had reached <code>CREATED</code> state in that environment.</p></body></html>
</body>
</html>

ichibahn:
<html>
<body>
<html><head></head><body><p>Refs #2329.</p>
<p><code>--regex</code> is evaluated against <code>database.table</code> for tables but against the <strong>bare database name</strong> for
the database object. A pattern like <code>'^mydb\.'</code> therefore selects every table in <code>mydb</code> while silently
excluding <code>mydb</code> itself, so <code>mydb-schema-create.sql</code> is never dumped — and mydumper still exits 0.</p>
<p>The resulting backup cannot be restored into a server where the database does not already exist
(<code>ERROR 1049</code>), and in <code>--stream</code> mode myloader holds every schema job until EOF, so the loader appears
to do nothing for the entire duration of the dump. Neither mydumper nor myloader says anything.</p>
<h3>What this changes</h3>
<p>Nothing functional — one warning, emitted once per database, only when all of the following hold:</p>
<ul>
<li><code>--regex</code> is in use (<code>is_regex_being_used()</code>)</li>
<li>the database object was <strong>not</strong> matched, so no <code>CREATE DATABASE</code> dump job was created</li>
<li>at least one table from that database <strong>is</strong> being dumped</li>
<li><code>--no-schemas</code> is not set</li>
</ul>
<blockquote>
<p>--regex matched tables in <code>mydb</code> but not the database object itself: mydb-schema-create.sql
will not be dumped, and this backup will not be restorable unless the database already exists
on the target. Database objects are matched against the bare database name, not
<code>database.table</code> - use for instance '^(mydb)(.|$)'.</p>
</blockquote>
<p><code>struct database</code> gains two <code>gboolean</code> flags, both initialised in <code>new_database()</code>:
<code>schema_create_job_created</code>, set in <code>get_database()</code> where the job is actually queued, and
<code>regex_mismatch_warned</code> as the one-time guard. The call site is immediately after the table-level
<code>eval_regex()</code> check in <code>dump_database_thread()</code> — the only path where the condition can arise.
<code>--tables-list</code> and <code>-B</code> both create the schema-create job without consulting <code>--regex</code>
(<code>get_database(conn, dt[0], TRUE)</code> and <code>get_database(conn, db_items[i], !no_schemas)</code> respectively),
so only the all-databases path can select tables while excluding the database object.</p>
<p>No new includes are needed — <code>no_schemas</code> comes from <code>mydumper_global.h</code> and <code>is_regex_being_used()</code>
is already used by <code>new_database()</code> in the same file.</p>
<p>The follow-up commit fixes two problems found while getting CI green. <code>new_database()</code> allocates with
<code>g_new</code> and initialises each field explicitly; the two new flags were left indeterminate, so when
<code>schema_create_job_created</code> read as non-zero the warning was silently suppressed — undefined behaviour
that would differ per build. And the test's <code>logfile=</code> pointed inside <code>outputdir</code>, which the harness
removes before each case; <code>set_verbose()</code> opens the log file before the output directory exists, so
mydumper aborted in <code>m_critical()</code> and every CI job that runs the e2e suite failed there.</p>
<h3>Test</h3>
<p><code>test/specific_40/</code> follows the existing <code>specific_31</code> layout — it greps the harness's own
<code>/tmp/test_mydumper.log.tmp</code> and sets neither <code>logfile=</code> nor <code>verbose=</code> — and asserts that with
<code>regex=^specific_40\.</code>:</p>
<ol>
<li>the table schema file is produced,</li>
<li><code>specific_40-schema-create.sql</code> is <strong>not</strong> produced,</li>
<li>the log contains the warning.</li>
</ol>
<p>Point 3 fails on <code>master</code> and passes with this change.</p>
<p>It deliberately sets no <code>database=</code>: with <code>-B</code> the schema-create job is created regardless of
<code>--regex</code>, which would both dump <code>specific_40-schema-create.sql</code> and suppress the warning.</p>
<h3>Verified no false positives</h3>
<p>Built from this branch and run against MySQL 8.4 with the same flags the harness applies
(<code>--checksum-all</code>, <code>--logfile</code>); the first row is the harness case itself:</p>

invocation | warning | -schema-create.sql
-- | -- | --
--regex '^specific_40\.', 1 table | shown once | not dumped
--regex '^specific_40\.', 3 tables | shown once (per-database guard holds) | not dumped
--regex '^specific_40(\.\|$)' | not shown | dumped
-B specific_40 (no regex) | not shown | dumped
--no-schemas --regex '^specific_40\.' | not shown | not dumped
-B specific_40 --regex '^specific_40\.' | not shown | dumped

<p>The last row is a deliberate limitation: with <code>-B</code> the schema-create job is created regardless of
<code>--regex</code>, so there is nothing to warn about.</p>
<p><code>-e 28:40 --skip-dynamic -d SSL</code> also passes, with <code>specific_40</code> running after the earlier cases have
left their databases on the server.</p>
<h3>Not included here</h3>
<p>Two further ideas from #2329, left out to keep this reviewable:</p>
<ul>
<li>the matching myloader-side warning, where schema jobs start being held in <code>schema_push()</code></li>
<li>documenting the two-subject matching rule in the <code>--regex</code> help text</li>
</ul>
<p>Happy to send either as a follow-up. I'd also welcome your view on the question in the issue — whether
myloader should fall back to <code>CREATE DATABASE IF NOT EXISTS</code> when no schema-create file arrives, rather
than holding every schema job until EOF.</p></body></html>
</body>
</html>

### DIFF
--- a/src/mydumper/mydumper_database.c
+++ b/src/mydumper/mydumper_database.c
@@ -56,6 +56,8 @@ static struct database *new_database(MYSQL *conn, char *database_name)
   _database->source_database_escaped = escape_string(conn, _database->source_database);
   //  _database->already_dumped = already_dumped;
   //  _database->ad_mutex=g_mutex_new();
+  _database->schema_create_job_created = FALSE;
+  _database->regex_mismatch_warned = FALSE;
   _database->checksum.schema = NULL;
   _database->checksum.routine = NULL;
   _database->checksum.trigger = NULL;
@@ -136,13 +138,14 @@ void warn_if_schema_create_excluded(struct database *database)
   if (!database->regex_mismatch_warned)
   {
     database->regex_mismatch_warned = TRUE;
-    g_warning("--regex matched tables in `%s` but not the database object itself: "
-              "%s-schema-create.sql will not be dumped, and this backup will not be "
-              "restorable unless the database already exists on the target. Database "
-              "objects are matched against the bare database name, not `database.table` "
-              "- use for instance '^(%s)(\\.|$)'.",
-              database->source_database, database->database_name_in_filename,
-              database->source_database);
+    g_warning(
+        "--regex matched tables in `%s` but not the database object itself: "
+        "%s-schema-create.sql will not be dumped, and this backup will not be "
+        "restorable unless the database already exists on the target. Database "
+        "objects are matched against the bare database name, not `database.table` "
+        "- use for instance '^(%s)(\\.|$)'.",
+        database->source_database, database->database_name_in_filename,
+        database->source_database);
   }
   g_mutex_unlock(database_hash_mutex);
 }
--- a/src/mydumper/mydumper_database.c
+++ b/src/mydumper/mydumper_database.c
@@ -109,10 +109,42 @@ struct database *get_database(MYSQL *conn, char *database_name, gboolean create_
   {
     database = new_database(conn, database_name);
     if (create_job)
+    {
       create_job_to_dump_schema(database);
+      database->schema_create_job_created = TRUE;
+    }
   }
   g_mutex_unlock(database_hash_mutex);
   return database;
+}
+
+/* Called when a table of `database` has been elected to be dumped.
+
+   --regex is evaluated against `database.table` for tables but against the bare
+   database name for the database object, so a pattern such as '^mydb\.' selects
+   every table while silently excluding the database itself.  The resulting
+   backup has no <db>-schema-create.sql and cannot be restored into a server
+   where the database does not already exist; in --stream mode myloader also
+   holds every schema job until EOF.  mydumper exits 0 either way, so warn once
+   per database instead of failing silently.  See issue #2329. */
+void warn_if_schema_create_excluded(struct database *database)
+{
+  if (no_schemas || !is_regex_being_used() || database->schema_create_job_created)
+    return;
+
+  g_mutex_lock(database_hash_mutex);
+  if (!database->regex_mismatch_warned)
+  {
+    database->regex_mismatch_warned = TRUE;
+    g_warning("--regex matched tables in `%s` but not the database object itself: "
+              "%s-schema-create.sql will not be dumped, and this backup will not be "
+              "restorable unless the database already exists on the target. Database "
+              "objects are matched against the bare database name, not `database.table` "
+              "- use for instance '^(%s)(\\.|$)'.",
+              database->source_database, database->database_name_in_filename,
+              database->source_database);
+  }
+  g_mutex_unlock(database_hash_mutex);
 }
 
 // see print_dbt_on_metadata_gstring() for table write to metadata
--- a/src/mydumper/mydumper_database.h
+++ b/src/mydumper/mydumper_database.h
@@ -31,10 +31,13 @@ struct database
   gchar                         *database_name_in_filename;
   struct database_level_checksum checksum;
   gboolean                       dump_triggers;
+  gboolean                       schema_create_job_created;
+  gboolean                       regex_mismatch_warned;
 };
 
 void             initialize_database();
 struct database *get_database(MYSQL *conn, char *database_name, gboolean create_job);
+void             warn_if_schema_create_excluded(struct database *database);
 void             free_databases();
 void             write_database_on_disk(FILE *mdfile);
 // OPTIMIZATION: Unsorted version for faster finalization
--- a/src/mydumper/mydumper_working_thread.c
+++ b/src/mydumper/mydumper_working_thread.c
@@ -389,6 +389,8 @@ static void get_table_info_to_process_from_list(MYSQL *conn, struct configuratio
       if (!eval_regex(database->source_database, row[0]))
         continue;
 
+      warn_if_schema_create_excluded(database);
+
       create_job_to_dump_table(is_view, is_sequence, database, g_strdup(row[tablecol]), g_strdup(row[collcol]), g_strdup(row[ecol]));
     }
     mysql_free_result(result);
@@ -1422,6 +1424,8 @@ static void dump_database_thread(MYSQL *conn, struct database *database)
       dump_summary_note_skipped();
       continue;
     }
+
+    warn_if_schema_create_excluded(database);
 
     /* Check if the table was recently updated */
     if (no_updated_tables && !is_view && !is_sequence)
--- a/src/mydumper/mydumper_working_thread.c
+++ b/src/mydumper/mydumper_working_thread.c
@@ -389,8 +389,6 @@ static void get_table_info_to_process_from_list(MYSQL *conn, struct configuratio
       if (!eval_regex(database->source_database, row[0]))
         continue;
 
-      warn_if_schema_create_excluded(database);
-
       create_job_to_dump_table(is_view, is_sequence, database, g_strdup(row[tablecol]), g_strdup(row[collcol]), g_strdup(row[ecol]));
     }
     mysql_free_result(result);
--- a/test/specific_40/check_mydumper.sh
+++ b/test/specific_40/check_mydumper.sh
@@ -1,0 +1,23 @@
+#!/usr/bin/env bash
+
+log=/tmp/data/specific_40.log
+
+# The tables must be dumped ...
+if [ ! -f /tmp/data/specific_40.t1-schema.sql ]
+then
+  exit 1
+fi
+
+# ... while the database object is not matched by '^specific_40\.' ...
+if [ -f /tmp/data/specific_40-schema-create.sql ]
+then
+  exit 1
+fi
+
+# ... and that must be reported instead of passing silently.
+if ! grep -q 'schema-create.sql will not be dumped' "$log"
+then
+  exit 1
+fi
+
+exit 0
--- a/test/specific_40/check_mydumper.sh
+++ b/test/specific_40/check_mydumper.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-log=/tmp/data/specific_40.log
+LOG=/tmp/test_mydumper.log.tmp
 
 # The tables must be dumped ...
 if [ ! -f /tmp/data/specific_40.t1-schema.sql ]
@@ -15,7 +15,7 @@ then
 fi
 
 # ... and that must be reported instead of passing silently.
-if ! grep -q 'schema-create.sql will not be dumped' "$log"
+if ! grep -q 'specific_40-schema-create.sql will not be dumped' "$LOG"
 then
   exit 1
 fi
--- a/test/specific_40/mydumper.cnf
+++ b/test/specific_40/mydumper.cnf
@@ -1,0 +1,12 @@
+#
+# A --regex that only matches `database.table` excludes the database object
+# itself, so <db>-schema-create.sql is not dumped. mydumper must warn about it
+# instead of exiting successfully with a backup that cannot be restored.
+# See issue #2329.
+#
+
+[mydumper]
+regex=^specific_40\.
+outputdir=/tmp/data
+logfile=/tmp/data/specific_40.log
+verbose=3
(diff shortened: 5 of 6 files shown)


davidducos:
Hi @ichibahn , seems that tests has failed, please review and fix it so I can proceed. Thanks!

ichibahn:
Hi @davidducos
I fixed in 860e55c. The test's `logfile=` pointed inside `outputdir`, which the
harness removes before each case, so mydumper aborted before it could open it. It now
greps the harness's own log, as `test/specific_31` does. Two smaller fixes came out of
verifying that: the new `struct database` flags were left uninitialised, and one call site
was unreachable. Details are in the updated PR description.

davidducos (approved):

PR #2335, opened 2026-08-27
Cut per-byte work in the row write path: fewer copies, fused escape pass

semx:
Two commits reducing the per-byte work in the row write path. On an 8-thread dump of a 1.7GB database (6.4M mixed rows plus 200k escape-heavy TEXT rows, MySQL 8.4) they cut client user CPU by about a third with byte-identical output:

| metric (7 runs) | INSERT | LOAD_DATA |
| --- | --- | --- |
| client user CPU, before | 6.3 s | 7.7 s |
| client user CPU, after | 4.2 s (-34%) | 4.9 s (-36%) |

Wall clock on my rig was server-bound (MySQL in a VM), so the wall gain there shows only when the client is the bottleneck; with `--threads 1` the patched build won all interleaved LOAD_DATA runs by ~5%. On setups where dump threads compete for CPU with compression or many parallel streams the CPU reduction is the point.

**Commit 1: fewer copies.** Every dumped byte travelled MySQL row -> escaped buffer -> column buffer -> row buffer -> statement buffer -> write(). Escaping and hex-encoding now write directly into the tail of the destination buffer, and when no masking functions are configured, columns go straight into the row buffer; the LOAD DATA fixup passes run in place. Also the global mutex protecting the `max-statement-size` statistic on every flush of every thread is replaced with a lock-free CAS maximum.

**Commit 2: one escape pass instead of three.** `mysql_real_escape_string()` walks input byte by byte through charset dispatch, and the LOAD DATA form then takes two more full passes (backslash -> escape char, terminator escaping). A table-driven single pass emits the same bytes directly. It is gated to where it is provably byte-identical to the library call:

- single-byte charsets (latin1, ascii, binary): any input;
- utf8 connections: non-binary columns only (server-validated contents); binary columns keep `mysql_real_escape_string()`, whose handling of invalid multibyte sequences differs;
- any other charset (gbk, sjis, ...): unchanged code path.

**Verification**

- Differential fuzz harness comparing old chain vs fused pass byte for byte: 50k random buffers per charset (arbitrary bytes for latin1/binary, generated valid utf8 incl. all escape-relevant ASCII for utf8mb4), five escape/terminator configurations for the LOAD DATA form including the pathological `esc == term`. No differences; the gate correctly refuses gbk. Happy to contribute the harness as a test if you want it in-tree.
- End-to-end: dump + myloader restore, `CHECKSUM TABLE` matches the source for both INSERT and LOAD_DATA formats.


davidducos:
Hi @semx , 
Did you test this for different charsets like latin1? I'm wondering if this will work on all the cases not just with UTF8...

semx:
Hi @davidducos, yes - that is exactly what the differential fuzz harness in the PR description covers. It compares the fused pass against the current chain (mysql_real_escape_string plus the two LOAD DATA fixup passes) byte for byte on a live connection, per charset:

- latin1 and binary connections: 50k random buffers of arbitrary bytes (all 256 values, lengths 0-2048), for the INSERT form and five LOAD DATA escape/terminator configurations each (including the pathological escaped-by == terminated-by). Byte-identical in all of them.
- utf8mb4: same volume with generated valid utf8 (ascii specials plus 2-, 3- and 4-byte sequences). Byte-identical. Binary columns on utf8 connections intentionally do not take the fused path (field.charsetnr == 63 falls back to mysql_real_escape_string), because for bytes that do not form valid utf8 the library escapes orphan bytes differently.
- gbk (and any other multibyte non-utf8 charset like sjis, big5): the gate refuses the fused path entirely and the code takes the unchanged mysql_real_escape_string route, precisely because trail bytes there can look like 0x5C.

So charsets fall in three buckets: single-byte (latin1/ascii/binary) fused for any input, utf8 fused for server-validated text columns only, everything else untouched. Happy to add the harness to the tree as a test, or extend the charset list in either direction if you prefer a narrower or wider gate.

davidducos (approved):

Issue #2183, opened 2026-02-05
[BUG] Performance problem introduced with version mydumper_0.13.1-1.jammy_amd64.deb gzip version


Nebucatnetzer:
**Describe the bug**

We were using mydumper v 0.10.1-1 from the Ubuntu 22.04 repositories for a long time. Our main database took around 20 minutes to dump.
Due to a problem with another system we tried the newest version of mydumper, that problem with the other system went away so we decided to roll out the upstream version across all servers for consistency.

However we quickly noticed that the performance with the new version is much worse as a dump now took around 1h 17min, which is move then four times as long as we are used to.
After some investigation that it was only bad as soon as compression is used.
Without compression the performance is slightly better than with 0.10.1-1 with gzip compression.

I did some tests with the various available builds and I think I narrowed it down to the build mydumper_0.13.1-1.jammy_amd64.deb. The zstd version of that build works fast and even quicker than 0.10.1-1.

mydumper_0.14.5-3-zstd seems to be the last version that had a distinct release for zstd and that build is still fast while the gzip version is slow. Every build afterwards is a combined build and they are all slow (granted I didn't test all of them).

You can see my test results including the commands I used in the following Google Sheet or the attached ODS.

- https://docs.google.com/spreadsheets/d/1P2dgjRYE8DoJLFyZc9x-Trk0E9t4BsR05E0sXNzcffI/edit?usp=sharing
- [mydummper performance.ods](https://github.com/user-attachments/files/25098167/mydummper.performance.ods)

Let me know if you need any additional information.

**To Reproduce**
Command executed:
* mydumper with all the parameters
* myloader with all the parameters

What mydumper and myloader version has been used?

**Expected behavior**

Performance should be on par with 0.10.1-1 or at least not four times slower.

**Log**

**Backup**

**How to repeat**

**Environment (please complete the following information):**
 - OS version: Ubuntu 22.04
 - MyDumper version: v0.21.3-1-v0.13.1-2 gzip version

**Additional context**

davidducos:
Hi @Nebucatnetzer , this is shocking, as I would not expect such a performance degradation. Is it possible for you to share 2 mydumper logs? from version v0.10 and the latest?

Nebucatnetzer:
I will try yes.

Nebucatnetzer:
I sent you an email to the address you have in your GitHub profile.
To keep this issue as open as possible I will mirror a part of the mail.


---

I used the following commands to create the logs:
**latest**
```
[...]

Issue #1889, opened 2025-05-28
Attempted to unlock an unlocked mutex?????


fukafuka224:
OS : Ubuntu 20.04
DB : MariaDB 10.11
Mydumper Version : mydumper v0.19.1-3, built against MySQL 8.0.41-32 with SSL support

When recovering with myloader, it logs the message "Attempted to unlock an unlocked mutex".

However, it confirms that all tables, data, routines, triggers, and views are recovered.

Commands used:
mydumper -u user -p password -i connect -t 8 -c --trx-tables -G -E -R -v 3 -W -o ${BACKUP_DIR} -B ${DATABASE}
myloader -u user -p password -t 8 -o -B ${DATABASE} -d ${BACKUP_DIR}

Is there a workaround?

**Describe the bug**

**To Reproduce**
Command executed:
* mydumper with all the parameters
* myloader with all the parameters

What mydumper and myloader version has been used?

**Expected behavior**

**Log**
If applicable, add the --verbose 4 / -v 4 and --logfile / -L <filename> to the execution of the command and upload the file.
Take into account that you can also use --masquerade-filename to do not expose database and table names.

**Backup**
If applicable, add the list of content of the database path, an `ls -l` will be enough.
Take into account that you can also use --masquerade-filename to do not expose the filenames.

**How to repeat**
If applicable, add the minimal table structure and data that we need to reproduce the issue.
Or upload a [core dump](https://github.com/mydumper/mydumper/wiki/Support#core-dump) for cases where mydumper or myloader is crashing.

**Environment (please complete the following information):**
 - OS version: [e.g. CentOS 7.9, Ubuntu 20.04]
 - MyDumper version: [ mydumper --version ]

**Additional context**

davidducos:
Hi @fukafuka224 , 
This looks like a bug, can you share the log of myloader? you should set -v 4 and -L to save it in a file.

fukafuka224:
Hi, @davidducos 
Thank you for your reply.

This is the result of running it with the -v 4 -L options added.
myloader -u user -p password -t 8 -o -B -v 4 -L test.log ${DATABASE} -d ${BACKUP_DIR} >> #{LOG_FILE} 2>&1

> test.log
2025-05-30 07:34:53 [INFO] - restoring index: Database.table
2025-05-30 07:34:53 [INFO] - Thread 23: restoring indexes Database.table from index. Tables 1597 of 1608 completed
[...]

PR #2338, opened 2026-09-01
myloader: remove fifo_hash entry on myl_close to stop stale FILE* lookups

DaviMedrade:
## Summary

Remove a file's `fifo_hash` entry in `myl_close()` so a later file that reuses the same `FILE*` address does not pick up the stale entry and abort myloader.

## Bug

`fifo_hash` maps `FILE*` → `struct fifo` for files read through a decompressor. `myl_close()` looked the entry up, `waitpid`ed the child and unlocked the entry's mutex, but never removed the entry.

`fclose()` frees the `FILE`, and a later `fopen()` can return the same address for a file that has no decompressor. The next `myl_close()` on that file then finds the stale entry and unlocks a mutex nobody holds, and glib aborts:

```
GLib: Attempt to unlock mutex that was not locked
```

Any restore mixing compressed and uncompressed files hits it — for instance a compressed dump where some files were decompressed by hand, or a stream carrying both.

## Fix

- `myl_close()`: remove the entry under `fifo_table_mutex` before `fclose()`, then free it after `waitpid()`.
- `myl_open()`: always create a fresh entry. The previous "entry already exists for this `FILE*`" branch only handled address reuse and cannot trigger once entries are removed on close.
- Drop `struct fifo.mutex`: it existed only to serialise that reuse, and was locked in one thread and unlocked in another, which GMutex does not allow.

## Test

Dump `--compress`, decompress every third file, restore from the directory with `-t 8`:

- Before: 10/10 runs abort with the message above.
- After: 10/10 complete; row and table counts match the source.

`ctest` passes; build is clean with `-Wall -Wextra -Werror`, on master at `26f2b295`.

### DIFF
--- a/src/myloader/myloader_process.c
+++ b/src/myloader/myloader_process.c
@@ -153,29 +153,14 @@ FILE *myl_open(char *filename, const char *type)
         m_remove(directory,tmpbasename);
         g_free(tmpbasename);
     */
+    struct fifo *f = g_new0(struct fifo, 1);
+    f->pid = child_proc;
+    f->filename = g_strdup(filename);
+    f->stdout_filename = fifoname;
+    f->uses_decompressor = TRUE;
     g_mutex_lock(fifo_table_mutex);
-    struct fifo *f = g_hash_table_lookup(fifo_hash, file);
-    if (f != NULL)
-    {
-      g_mutex_lock(f->mutex);
-      g_mutex_unlock(fifo_table_mutex);
-      f->pid = child_proc;
-      f->filename = g_strdup(filename);
-      f->stdout_filename = fifoname;
-      f->uses_decompressor = TRUE;
-    }
-    else
-    {
-      f = g_new0(struct fifo, 1);
-      f->mutex = g_mutex_new();
-      g_mutex_lock(f->mutex);
-      f->pid = child_proc;
-      f->filename = g_strdup(filename);
-      f->stdout_filename = fifoname;
-      f->uses_decompressor = TRUE;
-      g_hash_table_insert(fifo_hash, file, f);
-      g_mutex_unlock(fifo_table_mutex);
-    }
+    g_hash_table_insert(fifo_hash, file, f);
+    g_mutex_unlock(fifo_table_mutex);
   }
   else
   {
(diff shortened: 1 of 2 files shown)


davidducos:
Hi @DaviMedrade , thank you for sharing the PR!! I need some clarification to understand the kind of issue that you are solving, do you have a simple test case that I can reproduce?
Thanks!

DaviMedrade:
This is #1889. That log shows the same thread closing `…-schema-post.sql.gz` and then `…-schema-triggers.sql`, and aborting there — the second `fclose`/`fopen` reused the `FILE*` address, so `myl_close()` found the first file's `fifo_hash` entry and unlocked its mutex a second time. Reproduced here with the same log shape.

mydumper itself compresses either every file or none, so a directory like that only comes from files being decompressed (or compressed, as in the `zstd --rm data/*.sql` suggestion in #2183) after the dump. myloader handles such a mix without issue (well, except for this one) — every file is dispatched on its own suffix, and with this fix a mixed directory restores with all checksums confirmed.

Reproducer (aborts 30/30, any `-t`):
```
mydumper -B sakila -o data --compress --clear
cd data && ls *.zst | awk 'NR%3==0' | xargs zstd -d --rm ; cd ..
myloader -d data -B sakila2 -o --max-threads-for-schema-creation 1
```
"Attempt to unlock mutex that was not locked", SIGABRT. With a single uncompressed file (e.g. an edited `-schema-post.sql`) it's ~1 in 10 — the freed `FILE` has to come back from tcache before the next `fopen` in that thread — which is why #1889 wasn't reproducible on demand.

The stale entry is also the `myl_open`:160 / `myl_close`:210 lock-order inversion glibguard reports on every compressed restore; with a single-arena allocator it deadlocks the schema workers (gdb: one thread holding `fifo_table_mutex` waiting on `f->mutex`, another the reverse). The per-thread cache in glibc, jemalloc and tcmalloc alike is what keeps that from happening in practice.

davidducos (approved):

maxgoldberg:
Was getting `Attempt to unlock mutex that was not locked` when using myloader on v1.0.6, but v1.0.8 seems to have fixed it! Thanks!

